PR #1, opened 2018-08-21
Initial commit

miroslavKovacPantheon:
Statistics are upgraded to use more memory
    but the process is speeded up. All the modules
    are loaded in the memory at the begining so we
    don t have to call a request with every single
    module.

    All the utility files are commited too.

Signed-off-by: miroslav.kovac <miroslav.kovac@pantheon.tech>


evyncke:
Thank you Miroslav, I appreciate the structure and the README.md and the comments in the code.

Can you add the Apache licence in all python scripts? Copyright "Cisco and its affiliates" and obviously Authors yourself + your Pantheon email (or at least affiliation) ?

evyncke:
Also, the config.ini can you rename into config.ini.example (and I like the fact that there is no real password in it)

PR #6, opened 2018-08-30
Add parse and populate tooling

miroslavKovacPantheon:
Signed-off-by: miroslav.kovac <miroslav.kovac@pantheon.tech>


miroslavKovacPantheon:
@evyncke This PR needs to be talked through. It is starting to be tricky here since this is dependent on api part of the code.

miroslavKovacPantheon:
@evyncke 
> can we use the PYANG lib directly rather than forking?
There is a way to do this apparently. But since we need it just for two cases which is 

- to get tree
- to get check-update-from output

we need to do little more coding for that. Could we just open an issue for that? other then that this patch should be fine

PR #100, opened 2019-10-07
add the test for the function add_all_untracked

stanislav-chlebec:
Signed-off-by: Stanislav Chlebec <stanislav.chlebec@pantheon.tech>

### DIFF
--- a/utility/repoutil.py
+++ b/utility/repoutil.py
@@ -92,14 +92,18 @@ class RepoUtil(object):
                     repo_temp.clone()
                     break
         if branch == 'master':
+            if path is not None:
+                repo_temp.remove()
             return repo_temp.repo.head.commit.hexsha
         else:
             refs = repo_temp.repo.refs
             for ref in refs:
                 if ref.name == branch or ref.name == 'origin/{}'.format(branch):
-                    return ref.commit.hexsha
-        if path is not None:
-            repo_temp.remove()
+                    foundhash = ref.commit.hexsha
+            if foundhash is not None:
+                if path is not None:
+                    repo_temp.remove()
+                return foundhash
         if self.logger is not None:
             self.logger.error('Git branch - {} - could not be resovled'.format(branch))
         return branch
--- a/utility/repoutil.py
+++ b/utility/repoutil.py
@@ -91,21 +91,21 @@ class RepoUtil(object):
                     repo_temp = RepoUtil(submodule._url)
                     repo_temp.clone()
                     break
+        found_hash = None
         if branch == 'master':
-            if path is not None:
-                repo_temp.remove()
-            return repo_temp.repo.head.commit.hexsha
+            found_hash = repo_temp.repo.head.commit.hexsha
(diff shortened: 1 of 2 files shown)


stanislav-chlebec:
I checked all code and fixed on my side
But this should be hardened in get_commit_hash function:
       def test_get_commit_hash(self):
...
                self.assertEqual(self.repo5.get_commit_hash('backend','tests'), 'd7d499416b443f2e0d7594cad0b13b551d90dd3a')
                self.assertEqual(self.repo5.get_commit_hash('bottle-yang-extractor-validator','tests'), 'd1ec44dbe8995f282355d5d30402c5e6fa13a477')
....

I see that you are cloning submodules in a temporary folder and then after hash is found you are trying to clean that temporary folder.
Actually it should be done before you return the found hash....
Please fix this:
              if ref.name == branch or ref.name == 'origin/{}'.format(branch):
                    return ref.commit.hexsha
I suggest to  Insert before return:
       if path is not None:
            repo_temp.remove()








> 
> 
> Could you please make one more test in here. Or test it in tear down. what makes more sense for you. When you call remove, it should delete temporary dir which is in /tmp/<some_hash>. we need to make sure that this is in fact gone. and there are no such files. I found that when calling get_commit_hash it has created one extra directory and was never deleted.
> 
> ```
>     def get_commit_hash(self, path=None, branch='master'):
>         self.updateSubmodule()
>         repo_temp = self
>         if path is not None:
>             for submodule in self.repo.submodules:
>                 if path in submodule.path:
>                     repo_temp = RepoUtil(submodule._url)
>                     repo_temp.clone()
>                     break
> ```
> 
> This is fixed now
> 
> ```
>         if path is not None:
>             repo_temp.remove()
> ```
> 
> but we need to make sure that no temporary directory will stay created.

miroslavKovacPantheon:
You are right. could you please provide fix with this commit? thank you

stanislav-chlebec:
> 
> 
> You are right. could you please provide fix with this commit? thank you

ok

utility/repoutil.py
```diff
@@ -92,14 +92,18 @@ def get_commit_hash(self, path=None, branch='master'):
...
             return repo_temp.repo.head.commit.hexsha
         else:
             refs = repo_temp.repo.refs
             for ref in refs:
                 if ref.name == branch or ref.name == 'origin/{}'.format(branch):
-                    return ref.commit.hexsha
-        if path is not None:
-            repo_temp.remove()
+                    foundhash = ref.commit.hexsha
+            if foundhash is not None:
```
miroslavKovacPantheon: found_hash must be initialized to None otherwise it will  say
`NameError: name 'found_hash' is not defined` if it will not be found in between refs
It can be set to None at the beginning of else statement on line 98
also please divide the name with _ so it is found_hash instead of foundhash

stanislav-chlebec: I fixed that and also refactored all code.

PR #126, opened 2020-07-16
Running scripts from admin-ui

SlavomirMazurPantheon:

### DIFF
--- a/api/receiver.py
+++ b/api/receiver.py
@@ -558,6 +558,8 @@ class Receiver:
                 final_response = self.run_ietf()
             elif body == 'reload_config':
                 self.load_config()
+            elif 'run_script' == arguments[0]:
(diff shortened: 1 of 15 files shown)


api/receiver.py
```diff
@@ -673,6 +675,27 @@ def start_receiving(self):
                 except Exception:
                     pass
 
+    def run_script(self, arguments):
+        module_name = arguments[0]
+        script_name = arguments[1]
+        body_input = json.loads(arguments[2])
+        try:
+            with open(self.temp_dir + "/run-script-api-stderr.txt", "w") as f:
```
miroslavKovacPantheon, line 683: I don t think we need to open this file here - it was used only when we were running the sub-process which is not this case

SlavomirMazurPantheon: File opening removed

PR #128, opened 2020-07-22
Ping endpoint to check if session is still valid

SlavomirMazurPantheon:


yang-catalog:
We should start using abort with description as in other modules but this needs to be changed in different patch through whole admin.py script. I am merging this

PR #135, opened 2020-08-14
Multiple log files filtering

SlavomirMazurPantheon:


api/views/admin/admin.py
```diff
@@ -26,7 +26,7 @@
 import re
 import shutil
 import stat
-from datetime import datetime
```
miroslavKovacPantheon, line 29: change the datetime back since we are not using it in here for sessions lifetime anymore

PR #174, opened 2021-01-22
Tests for capability.py

SlavomirMazurPantheon:
- Updated capability.py file
- Changed way of loading YangModels/yang repo - loaded from disk instead of cloning everytime
- Added docstrings all tests and capability.py script methods
- Travis prepare steps moved to separate prepare_environment.sh script
- Tests covering capability.py added
- Removed files which can be created during Travis before_script phase


tests/test_capability.py
```diff
@@ -0,0 +1,287 @@
...
+                                None, api, sdo, json_dir, yc_gc.result_dir,
+                                yc_gc.save_file_dir, yc_gc.private_dir, yc_gc.yang_models)
+
+        capability.parse_and_dump_sdo(repo)
+
+        with open('{}/prepare-sdo.json'.format(json_dir), 'r') as f:
+            sdos_json = json.load(f)
+
+        sdos_list = sdos_json['modules']['module']
+        for sdo in sdos_list:
```
miroslavKovacPantheon, line 100: test if the length is non zero as well so we don t end up with valid test and nothing in module list.. same in all cases

miroslavKovacPantheon (changes requested):

parseAndPopulate/capability.py
```diff
@@ -126,51 +132,54 @@ def __init__(self, log_directory: str, hello_message_file: str, prepare: Prepare
...
-                repo = repoutil.RepoUtil('{}{}/{}'.format(github_url, self.owner, self.repo), self.logger)
-                repo.clone()
+                repo_url = '{}{}/{}'.format(github_url, self.owner, self.repo)
+                repo = repoutil.load(self.yang_models_dir, repo_url)
                 self.path = None
                 self.branch = 'master'
                 self.branch = repo.get_commit_hash(self.branch)
                 repo.remove()
                 # Solve for os-type
-                if 'nx' in self.split[4]:
```
miroslavKovacPantheon, line 154: why did we remove split[4]. please show me or remind me what split holds. if I remember correctly the string itself could contain more then just nx in split[4] that is why I used 'nx' in self.split[4].

your way we are just looking if the list contains string 'nx' exactly as is

parseAndPopulate/capability.py
```diff
@@ -179,15 +188,19 @@ def __init__(self, log_directory: str, hello_message_file: str, prepare: Prepare
...
-            LOGGER.info('Parsing a received json file')
+            LOGGER.info('Parsing a received platform-metadata.json file')
             self.owner = impl['module-list-file']['owner']
             self.repo = impl['module-list-file']['repository'].split('.')[0]
             self.path = impl['module-list-file']['path']
             self.branch = impl['module-list-file'].get('branch')
-            repo = repoutil.RepoUtil('{}{}/{}'.format(github_url, self.owner, self.repo), self.logger)
-            repo.clone()
+            repo_url = '{}{}/{}'.format(github_url, self.owner, self.repo)
+            if self.owner == 'YangModels' and self.repo == 'yang':
```
miroslavKovacPantheon, line 199: did you test this? FYI nice touch

parseAndPopulate/capability.py
```diff
@@ -200,18 +213,24 @@ def initialize(self, impl):
...
             LOGGER.debug('Parsing sdo files sent via API')
-            with open(self.json_dir + '/prepare-sdo.json', 'r') as f:
+            with open('{}/prepare-sdo.json'.format(self.json_dir), 'r') as f:
                 sdos_json = json.load(f)
             sdos_list = sdos_json['modules']['module']
             for sdo in sdos_list:
                 file_name = unicodedata.normalize('NFKD', sdo['source-file']['path'].split('/')[-1]) \
                     .encode('ascii', 'ignore')
-                LOGGER.info('Parsing sdo file sent via API {}'.format(file_name))
+                LOGGER.info('Parsing sdo file sent via API {}'.format(file_name.decode('utf-8', 'strict')))
```
miroslavKovacPantheon, line 233: did this cause troubles before with no decode?

parseAndPopulate/capability.py
SlavomirMazurPantheon: self.split holds splitted path to hello_message_file by '/' character
I think that expected path is something like yangmodels/yang/vendors/cisco/nx/ ...
However, if (for example) absolute path is passed, then 'nx' wil not be on split[4]
So if we are able to get position of 'nx' (= platform name), then we can get indexes of vendors, software-type ...

parseAndPopulate/capability.py
SlavomirMazurPantheon: The message was in the following format:
Parsing sdo file sent via API b'name-of-module.yang'

miroslavKovacPantheon (approved):

PR #213, opened 2021-02-19
Changed crontab emails

SlavomirMazurPantheon:
- Emails removed from crontab, to avoid sending an emails during development


miroslavKovacPantheon:
not valid anymore

Issue #226, opened 2021-03-04
ConfD healthcheck - full check


SlavomirMazurPantheon:
Currently there is an endpoint that will only try to get the module from ConfD database:
```
api/admin/healthcheck/confd
```
Create new healthcheck endpoint which will test sending requests to ConfD database.
Make GET, PATCH, DELETE requests with non-existing module.
Results should be displayed in admin_ui.

PR #227, opened 2021-03-04
Healthcheck update - ConfD and Yangcatalog domains checks

### LINKED ISSUE #225: Yangcatalog domains check
Add healthcheck endpoint which will test the availability (send GET requests) of Yangcatalog domains.
Results should be displayed in admin_ui.

**List of URLs to check:**
- http://yangcatalog.org
- http://www.yangcatalog.org
- https://yangcatalog.org
- https://www.yangcatalog.org
- http://yangvalidator.com
- http://www.yangvalidator.com
- https://yangvalidator.com
- https://www.yangvalidator.com
- http://18.224.127.129
- https://18.224.127.129
- http://[2600:1f16:ba:200:a10d:3212:e763:e720]
- https://[2600:1f16:ba:200:a10d:3212:e763:e720]

SlavomirMazurPantheon:
- Created endpoint which tests the availability of Yangcatalog domains
- Creates healthcheck endpoint which tests sending requests to ConfD database

This resolves #225 #226 

### DIFF
--- a/api/views/healthCheck/healthCheck.py
+++ b/api/views/healthCheck/healthCheck.py
@@ -46,8 +46,9 @@ app = HealthcheckBlueprint('healthcheck', __name__)
 @app.route('/services-list', methods=['GET'])
 def get_services_list():
     response_body = []
-    service_endpoints = ['my-sql', 'elk', 'confd', 'yang-search-admin', 'yang-validator-admin', 'yangre-admin', 'nginx', 'rabbitmq']
-    service_names = ['MySQL', 'Elasticsearch', 'ConfD', 'YANG search', 'YANG validator', 'YANGre', 'NGINX', 'RabbitMQ']
+    service_endpoints = ['my-sql', 'elk', 'confd', 'yang-search-admin', 'yang-validator-admin',
+                         'yangre-admin', 'nginx', 'rabbitmq', 'yangcatalog', 'confd-full-check']
+    service_names = ['MySQL', 'Elasticsearch', 'ConfD', 'YANG search', 'YANG validator', 'YANGre', 'NGINX', 'RabbitMQ', 'YangCatalog', 'ConfD - Full check']
     for name, endpoint in zip(service_names, service_endpoints):
         pair = {'name': name, 'endpoint': endpoint}
         response_body.append(pair)
@@ -58,7 +59,7 @@ def get_services_list():
 def health_check_mysql():
     try:
         db = MySQLdb.connect(host=yc_gc.dbHost, db=yc_gc.dbName, user=yc_gc.dbUser,
-                                passwd=yc_gc.dbPass)
+                             passwd=yc_gc.dbPass)
         if db is not None:
             app.LOGGER.info('Successfully connected to database: {}'.format(yc_gc.dbName))
             # prepare a cursor object using cursor() method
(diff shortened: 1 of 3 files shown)


api/views/healthCheck/healthCheck.py
```diff
@@ -403,6 +404,147 @@ def health_check_yang_search_admin():
...
+            'https://www.yangcatalog.org',
+            'http://yangvalidator.com',
+            'http://www.yangvalidator.com',
+            'https://yangvalidator.com',
+            'https://www.yangvalidator.com',
+            'http://18.224.127.129',
+            'https://18.224.127.129',
+            'http://[2600:1f16:ba:200:a10d:3212:e763:e720]',
+            'https://[2600:1f16:ba:200:a10d:3212:e763:e720]'
+            ]
```
miroslavKovacPantheon, line 426: make it list of tuples or dictionary that will contain url and boolean for request verify variable

api/views/healthCheck/healthCheck.py
```diff
@@ -403,6 +404,147 @@ def health_check_yang_search_admin():
...
+
+    for url in urls:
+        result = {}
+        result['label'] = url
+        try:
+            if any(char.isdigit() for char in url):
+                response = requests.get(url, verify=False)
+            else:
+                response = requests.get(url)
```
miroslavKovacPantheon, line 435: use verify variable from tuple. url can change and be called something like yangcatalog2 and it s not safe that way.

api/views/healthCheck/healthCheck.py
```diff
@@ -403,6 +404,147 @@ def health_check_yang_search_admin():
...
+
+@app.route('/confd-full-check', methods=['GET'])
+def health_check_confd_full_check():
+    service_name = 'ConfD - Full check'
+    status = 'running'
+    message = 'All requests responded with status code 200'
+    additional_info = []
+    check_module_name = 'confd-full-check'
+    confd_prefix = '{}://{}:{}'.format(yc_gc.protocol, yc_gc.confd_ip, repr(yc_gc.confdPort))
+    headers = {'Content-type': 'application/yang-data+json', 'Accept': 'application/yang-data+json'}
```
miroslavKovacPantheon, line 459: please create a python file that will contain static variables
confd_header_str = 'applicaton/yang-data+json'
confd_content_type = {'Content-type': confd_header_str }
confd_accept = {'Accept': confd_header_str }
confd_headers = {confd_content_type , confd_accept }

and reuse confd_headers accross the whole project. We will update this file with any string or dict that is used multiple times so we have to do the change only on one place instead of multiple if something changes. I remember that confd headers already did once.

api/views/healthCheck/healthCheck.py
```diff
@@ -403,6 +404,147 @@ def health_check_yang_search_admin():
...
+
+    try:
+        # GET
+        result = {}
+        result['label'] = 'GET yang-catalog@2018-04-03'
```
miroslavKovacPantheon, line 464: I don t like the idea of playing with something that already exists in our database and accidentally removing it. especially with latest yang-catalog file that is used with every search. Although we are using redis I still want this database to be clean. you can create some json file output and save it into project out of this file but change name revision, organization and namespace. make first PUT request and make sure it is in there and continue from there as you did here with yang-catalog.

miroslavKovacPantheon (changes requested):

api/views/healthCheck/healthCheck.py
SlavomirMazurPantheon: yang-catalog@2018-04-03 is essential module which need to be in ConfD in first place, this is the reason why I am making GET of this module ... then I change name, which is basically same as loading some fictional module from json and using it for next checks ... However it is not a problem to move this to some json file and loading from it.

api/views/healthCheck/healthCheck.py
SlavomirMazurPantheon: done

api/views/healthCheck/healthCheck.py
SlavomirMazurPantheon: done

api/views/healthCheck/healthCheck.py
SlavomirMazurPantheon: done for healthCheck.py, rest will be added in another pull request

api/views/healthCheck/healthCheck.py
miroslavKovacPantheon: Oh sorry, I have missed that. You are right this is ok.

PR #249, opened 2021-04-08
ietfYangDraftPull module scripts refactored

### LINKED ISSUE #248: Refactor draftPullLocal.py
**Current state:**
Populate script is called using subprocess library. 
If an error occurs while the program is running, it is written to a .txt file.
However, this file is overwritten each time the draftPullLocal script is run, so this information may be lost before it is even checked.
Also, YangModels/yang repository is always cloned into a new folder (randomly generated suffix) making it impossible to use file [content hashing feature](https://github.com/YangCatalog/backend/issues/183).

**Possible improvement:**
- Change this to use ScriptConf of populate script and set all the necessary arguments.
- If an error occurs while the program is running, it will be logged to the appropriate log file.
- During this refactoring, change the directory where YangModels/yang repo is cloned. It should always have the same name.

SlavomirMazurPantheon:
- Changed the way how populate.py script is called
- Changed the way how draftPull.py script is called
- Shared methods moved into separate draftPullUtility.py script

This resolves #248 

### DIFF
--- a/ietfYangDraftPull/draftPullLocal.py
+++ b/ietfYangDraftPull/draftPullLocal.py
@@ -93,7 +93,7 @@ def run_populate_script(directory: str, notify: bool, LOGGER):
         script_conf.args.__setattr__('notify_indexing', notify)
         submodule.main(scriptConf=script_conf)
     except Exception as e:
-        LOGGER.error('populate script error:\n{}'.format(e))
+        LOGGER.exception('populate script error:\n{}'.format(e))
         successful = False
 
     return successful
@@ -179,7 +179,7 @@ def main(scriptConf=None):
             message = {'label': 'Experimental modules', 'message': 'populate script finished successfully'}
             messages.append(message)
     except Exception as e:
-        LOGGER.error('Exception found while running draftPullLocal script')
+        LOGGER.exception('Exception found while running draftPullLocal script')
         job_log(start_time, temp_dir, error=str(e), status='Fail', filename=os.path.basename(__file__))
         repo.remove()
         raise e
--- a/ietfYangDraftPull/draftPullLocal.py
+++ b/ietfYangDraftPull/draftPullLocal.py
@@ -26,22 +26,25 @@ __email__ = "miroslav.kovac@pantheon.tech"
 
 import argparse
 import os
-import subprocess
+import shutil
 import sys
-import tarfile
 import time
-from datetime import datetime
 
 import requests
-
 import utility.log as log
-from utility import repoutil, yangParser
-from utility.util import get_curr_dir, job_log
+from utility import repoutil
+from utility.util import job_log
 
 if sys.version_info >= (3, 4):
     import configparser as ConfigParser
 else:
     import ConfigParser
+
+from ietfYangDraftPull.draftPullUtility import (check_early_revisions,
+                                                check_name_no_revision_exist,
+                                                extract_rfc_tgz,
+                                                get_draft_module_content)
+
 
 class ScriptConfig:
     def __init__(self):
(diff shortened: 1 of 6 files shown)


ietfYangDraftPull/draftPullLocal.py
```diff
@@ -72,96 +75,28 @@ def get_help(self):
...
+    try:
+        module = __import__('parseAndPopulate', fromlist=['populate'])
+        submodule = getattr(module, 'populate')
+        script_conf = submodule.ScriptConfig()
+        script_conf.args.__setattr__('sdo', True)
+        script_conf.args.__setattr__('dir', directory)
+        script_conf.args.__setattr__('notify_indexing', notify)
+        submodule.main(scriptConf=script_conf)
+    except Exception as e:
+        LOGGER.error('populate script error:\n{}'.format(e))
```
miroslavKovacPantheon, line 96: LOGGER.exception instead of error. This will give us more information and traceback. https://docs.python.org/3/library/logging.html

you should be using that in any
 ```
except  Exception as e:
    ...
```
it should not be used outside of except clause

File: ietfYangDraftPull/draftPullUtility.py [hunk 2]
```diff
@@ -0,0 +1,186 @@
...
+                        if revision is None:
+                            continue
+                    try:
+                        # Basic date extraction can fail if there are alphanumeric characters in the revision filename part
+                        year = int(revision.split('-')[0])
+                        month = int(revision.split('-')[1])
+                        day = int(revision.split('-')[2])
+                        revisions.append(datetime(year, month, day))
+                    except Exception:
+                        LOGGER.error('Failed to process revision for {}: (rev: {})'.format(f2, revision))
```
miroslavKovacPantheon, line 125: same as above

ietfYangDraftPull/draftPullUtility.py
```diff
@@ -0,0 +1,186 @@
...
+        ietf_draft_json = response.json()
+    for key in ietf_draft_json:
+        with open('{}/{}'.format(experimental_path, key), 'w+') as yang_file:
+            yang_download_link = ietf_draft_json[key][2].split('href="')[1].split('">Download')[0]
+            yang_download_link = yang_download_link.replace('new.yangcatalog.org', 'yangcatalog.org')
+            try:
+                yang_raw = requests.get(yang_download_link).text
+                yang_file.write(yang_raw)
+            except:
+                LOGGER.warning('{} - {}'.format(key, yang_download_link))
```
miroslavKovacPantheon, line 162: This one is ok since we don t care about the error in here and we just move on. exception will log it as error and we want only warning in here.

utility/util.py
```diff
@@ -212,9 +213,8 @@ def prepare_to_indexing(yc_api_prefix: str, modules_to_index, credentials: list,
                                                          module['name'],
                                                          module['revision'],
                                                          module['organization'])
-                response = requests.get(url, auth=(credentials[0], credentials[1]),
-                                        headers={'Content-Type': 'application/json',
-                                                 'Accept': 'application/json'})
+                response = requests.get(url, headers={'Content-Type': 'application/json',
```
miroslavKovacPantheon, line 216: create json_headers and replace only in scripts that you are changing in this patch

miroslavKovacPantheon (changes requested):

ietfYangDraftPull/draftPullLocal.py
SlavomirMazurPantheon: Done

ietfYangDraftPull/draftPullUtility.py
SlavomirMazurPantheon: done

utility/util.py
SlavomirMazurPantheon: done - separate PR will be created for replacing this in rest of the scripts

miroslavKovacPantheon (approved):

PR #268, opened 2021-05-03
Sandbox script to revise tree types for all the modules

### LINKED ISSUE #259: YANG module metadata tree-type is wrong for older revisions
Hi,

It seems that the older revision of a YANG module still contain the metadata for the latest revision.
I'm looking at https://yangcatalog.org/yang-search/module_details/ietf-interfaces?module=ietf-interfaces, the latest 2018-02-20, for which the tree-type is nmda-compatible.
Now, if I take an older revision, e.g. https://yangcatalog.org/yang-search/module_details/ietf-interfaces?module=ietf-interfaces@2010-12-08 ... the tree-type is still nmda-compatible

Regards, Benoit

SlavomirMazurPantheon:
After running this script, tree types for all the modules will be checked/fixed - closes  #259 

### DIFF
--- a/sandbox/revise_tree_type.py
+++ b/sandbox/revise_tree_type.py
@@ -1,0 +1,64 @@
+import configparser as ConfigParser
+import json
+import sys
+import time
+from datetime import datetime
+
+import requests
+from parseAndPopulate.modulesComplicatedAlgorithms import \
+    ModulesComplicatedAlgorithms
+
+if __name__ == '__main__':
+    start = time.time()
+    config_path = '/etc/yangcatalog/yangcatalog.conf'
+    config = ConfigParser.ConfigParser()
+    config._interpolation = ConfigParser.ExtendedInterpolation()
+    config.read(config_path)
+    api_protocol = config.get('General-Section', 'protocol-api', fallback='http')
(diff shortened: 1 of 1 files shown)


sandbox/revise_tree_type.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # Initialize ModulesComplicatedAlgorithms
+    confd_prefix = '{}://{}:{}'.format(confd_protocol, confd_ip, repr(confd_port))
+    direc = '/var/yang/tmp'
+
+    num_of_modules = len(all_existing_modules['module'])
+    chunk_size = 100
+    chunks = (num_of_modules - 1) // chunk_size + 1
+    for i in range(chunks):
+        batch = all_existing_modules['module'][i * chunk_size:(i + 1) * chunk_size]
```
miroslavKovacPantheon, line 50: print i variable out of chunks variable. so if it fails we can just change that number and continue where we failed.

miroslavKovacPantheon (changes requested):

sandbox/revise_tree_type.py
SlavomirMazurPantheon: print added

miroslavKovacPantheon (approved):

PR #278, opened 2021-06-07
Check for different files with identical name and revision

richardzilincikPantheon:
sorry, not sure what the top path should be


sandbox/find_conflicting_files.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+conflicting = []
+fnames = {}
+top = '/backend/tests/resources/yangmodels/yang'
+for dirname, _, files in os.walk(top):
+    dirname = os.path.join(top, dirname)
+    for f in files:
+        if f.endswith('.yang'):
+            revision = find_revision(os.path.join(dirname, f))
```
miroslavKovacPantheon, line 24: This doesn t make it for sure that the revision is the last one. If you check the RFC it is said that each yang file can have several revisions as the yang file is updated.. it should be in order from newer to older but there are many cases that this is not true. 

In your find revision case you find the first one and return it. the parser that I proposed takes care of such cases that is why it s better to use that.

miroslavKovacPantheon (changes requested):

sandbox/find_conflicting_files.py
richardzilincikPantheon: The parser was consuming >7GB of RAM and freezing up my machine when I tried it. I'll try and get the latest revision with re if that's ok.

miroslavKovacPantheon: on which module was it?

make sure it s not just building up the schema context but you keep creating a new one.. otherwise with each module you are taking more and more memory..

richardzilincikPantheon: oh I see, how would I prevent this?

richardzilincikPantheon:
not sure where exactly the issue was, works now

tests/test_api_search.py
```diff
@@ -368,7 +368,7 @@ def test_get_organizations_no_modules(self, mock_redis_get: mock.MagicMock):
         self.assertEqual(len(contributors_list), 0)
 
     def test_create_update_from(self):
-        """Test if payload contains desired output = comparasion of pyang outputs for two different yang module revisions.
+        """Test if payload contains desired output = comparasion of pyang outputs for two eerent yang module revisions.
```
miroslavKovacPantheon, line 371: revert this

tests/test_api_search.py
```diff
@@ -403,6 +403,43 @@ def test_create_update_from_raise_exception(self, mock_os_makedirs: mock.MagicMo
...
+        revision1 = '2018-04-03'
+        file2 = 'yang-catalog'
+        revision2 = '2017-09-26'
+
+        path = 'api/services/diff-tree/file1={}@{}/file2={}@{}'.format(file1, revision1, file2, revision2)
+        result = self.client.get(path)
+        data = result.data.decode()
+
+        self.assertEqual(data, 'Server error - could not create directory')
+
```
miroslavKovacPantheon, line 442: please separate these changes in different PR. You need to create a new branch for yourself and out of that branch create a separate PR.

miroslavKovacPantheon (changes requested):

tests/test_api_search.py
richardzilincikPantheon: sorry, I was having trouble with git, done should be done now

SlavomirMazurPantheon (approved):

miroslavKovacPantheon (approved):

PR #288, opened 2021-06-24
Add backup script for mariadb

richardzilincikPantheon:
Perhaps the backups should automatically go into a specific directory? The /var/yang/cache directory is already used by recovery.py and the naming would clash.


api/views/admin/admin.py
```diff
@@ -787,7 +787,7 @@ def get_module_name(script_name):
         return 'parseAndPopulate'
     elif script_name == 'draftPull' or script_name == 'draftPullLocal' or script_name == 'openconfigPullLocal':
         return 'ietfYangDraftPull'
-    elif script_name == 'recovery' or script_name == 'elkRecovery' or script_name == 'elkFill':
+    elif script_name in ['recovery', 'elkRecovery', 'elkFill', 'mariadbRecovery']:
```
SlavomirMazurPantheon, line 790: nice catch - please use this syntax also above for draftPull ...

recovery/mariadbRecovery.py
```diff
@@ -0,0 +1,112 @@
...
+
+def main(scriptConf=None):
+    if scriptConf is None:
+        scriptConf = ScriptConfig()
+    args = scriptConf.args
+    config = ConfigParser.ConfigParser()
+    config._interpolation = ConfigParser.ExtendedInterpolation()
+    config.read(args.config_path)
+    log_directory = config.get('Directory-Section', 'logs')
+    backup_directory = '{}/mariadb'.format(config.get('Directory-Section', 'cache'))
```
SlavomirMazurPantheon, line 87: please assign result of config.get() to variable named cache_dir or cache_directory and use format later
It is just a detail, but we want this to be same through all the scripts for easier identification in case of changes

api/views/admin/admin.py
```diff
@@ -783,11 +783,11 @@ def get_disk_usage():
 
 ### HELPER DEFINITIONS ###
 def get_module_name(script_name):
-    if script_name == 'populate' or script_name == 'runCapabilities':
+    if script_name in ['pupulate', 'runCapabilities']:
```
SlavomirMazurPantheon, line 786: typo in populate

SlavomirMazurPantheon (approved):

PR #291, opened 2021-06-25
Remove duplicate imports

richardzilincikPantheon:


api/views/userSpecificModuleMaintenace/moduleMaintanace.py
```diff
@@ -34,8 +34,6 @@
 from api.authentication.auth import auth, hash_pw
 
 from api.globalConfig import yc_gc
-from flask import Blueprint, abort, jsonify, make_response, request
```
miroslavKovacPantheon, line 37: Why is it ok to get rid of these?

They are used in this file and will be missing if you remove that import. I don t see the duplicity.

miroslavKovacPantheon (changes requested):

api/views/userSpecificModuleMaintenace/moduleMaintanace.py
richardzilincikPantheon: they are already imported on line 31

miroslavKovacPantheon (approved):

api/views/userSpecificModuleMaintenace/moduleMaintanace.py
miroslavKovacPantheon: you are right

PR #302, opened 2021-07-13
Add tests for ycJobs.py

richardzilincikPantheon:


SlavomirMazurPantheon:
What about api/ietf endpoint? 
https://github.com/YangCatalog/backend/blob/3db6227e3adea7af90db565720f1e96e3a58b0d6/api/views/ycJobs/ycJobs.py#L43
Is it possible to create tests also for this one?

richardzilincikPantheon:
I couldn't think of a way to bypass the authentication because I thought a mock wouldn't work with the authentication decorators, but I realized the decorators let me to overwrite the previous callback.

PR #303, opened 2021-07-13
Simplify returns and clean up in ycJobs.py

richardzilincikPantheon:


SlavomirMazurPantheon:
Please reverse order of those two decorators: https://github.com/YangCatalog/backend/blob/3db6227e3adea7af90db565720f1e96e3a58b0d6/api/views/ycJobs/ycJobs.py#L42
Route first, then login_required

PR #306, opened 2021-07-16
Make reviseSemver runnable from the admin-ui

richardzilincikPantheon:


SlavomirMazurPantheon:
Please add name of this script to admin.py script so it will be possible to run this from admin.

PR #308, opened 2021-07-23
Add reviseTreeType script and a cronjob for it

richardzilincikPantheon:


SlavomirMazurPantheon:
Some logs would be useful, for example:
- Starting reviseTreeType script
- Job finished successfully
- X modules sent for tree type resolving
...

and log them into parseAndPopulate.log log file

SlavomirMazurPantheon (approved):

PR #334, opened 2021-08-11
Load config path from env var

richardzilincikPantheon:


SlavomirMazurPantheon:
Did you tried to build it locally? There are problems with two containers - yc-api-receiver and yc-api-recovery.
This need to be added also to api-receiver and api-recovery in docker-compose.yml file:
https://github.com/YangCatalog/deployment/blob/a3aa0f0f7e49070900ee795b70fefbc92fed1010/docker-compose.yml#L28

Otherwise they do not have YANGCATALOG_CONFIG_PATH as env variable.

richardzilincikPantheon:
My bad, I thought I did, but I must have forgotten to save a file before building or something, I see the issue now.

SlavomirMazurPantheon:
Also it seems like there is a problem if script is run as cronjob. 
It is absolutely ok if I run script from console for example, but getting this error if script is run as cronjob.
```
  File "/backend/elasticsearchIndexing/process-changed-mods.py", line 26, in <module>
    from elasticsearchIndexing import build_yindex
  File "/backend/elasticsearchIndexing/build_yindex.py", line 39, in <module>
    from utility.util import fetch_module_by_schema, get_curr_dir
  File "/backend/utility/util.py", line 492, in <module>
    def create_config(config_path=os.environ['YANGCATALOG_CONFIG_PATH']):
  File "/usr/local/lib/python3.9/os.py", line 679, in __getitem__
    raise KeyError(key) from None
KeyError: 'YANGCATALOG_CONFIG_PATH'
```
Please, try to check this too.

SlavomirMazurPantheon:
Path to the config need to be passed here:
https://github.com/YangCatalog/backend/blob/3555b53135b776b7c32d480963fe613b76f4d094/api/views/admin/admin.py#L227
https://github.com/YangCatalog/backend/blob/3555b53135b776b7c32d480963fe613b76f4d094/api/views/admin/admin.py#L241
Please, use os.environ here too.

utility/util.py
```diff
@@ -408,10 +407,7 @@ def get_module_from_es(name: str, revision: str):
         :param name         (str) name of the module
         :param revision     (str) revision of the module in format YYYY-MM-DD
     """
-    config_path = '/etc/yangcatalog/yangcatalog.conf'
-    config = ConfigParser.ConfigParser()
-    config._interpolation = ConfigParser.ExtendedInterpolation()
-    config.read(config_path)
+    config = create_config()
```
SlavomirMazurPantheon, line 410:
```
````
import is missing

PR #361, opened 2021-09-21
Add registration time and motivation for users

richardzilincikPantheon:
The `/api/admin/move-user` and `/api/admin/sql-tables/<table>` endpoints now also check for a `motivation` field. The `/api/admin/sql-tables/<table>` now also returns `motivation` and `registeration-datetime` fields. `/register-user` now expects a `motivation` field.
Before this code is updated, the database should be modified with the following commands:
```sql
ALTER TABLE users
ADD COLUMN RegistrationDatetime TIMESTAMP DEFAULT CURRENT_TIMESTAMP;
ALTER TABLE users_temp
ADD COLUMN Motivation VARCHAR(255)
ADD COLUMN RegistrationDatetime TIMESTAMP DEFAULT CURRENT_TIMESTAMP;
```

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -17,7 +17,7 @@ __copyright__ = "Copyright The IETF Trust 2021, All Rights Reserved"
 __license__ = "Apache License, Version 2.0"
 __email__ = "richard.zilincik@pantheon.tech"
 
-from sqlalchemy import Column, Integer, String, DateTime
+from sqlalchemy import Column, Integer, String, DateTime, text
 from sqlalchemy.ext.declarative import DeferredReflection
 from sqlalchemy.orm import declarative_base
 
@@ -36,7 +36,6 @@ class BaseUser(DeferredReflection, Base):
     AccessRightsSdo = Column(String(255))
     AccessRightsVendor = Column(String(255))
     Motivation = Column(String(255))
-    RegistrationDatetime = Column(DateTime())
 
 
 class User(BaseUser):
@@ -45,3 +44,4 @@ class User(BaseUser):
 
 class TempUser(BaseUser):
     __tablename__ = 'users_temp'
+    RegistrationDatetime = Column(DateTime)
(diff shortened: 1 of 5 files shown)


api/views/admin/admin.py
```diff
@@ -563,7 +570,9 @@ def get_sql_rows(table):
                     'first-name': user.FirstName,
                     'last-name': user.LastName,
                     'access-rights-sdo': user.AccessRightsSdo,
-                    'access-rights-vendor': user.AccessRightsVendor}
+                    'access-rights-vendor': user.AccessRightsVendor,
+                    'motivation': user.Motivation,
+                    'registeration-datetime': user.RegisterationDatetime}
```
SlavomirMazurPantheon, line 575: There is a type: **user.RegisterationDatetime** -> **user.RegistrationDatetime**

api/views/userSpecificModuleMaintenance/moduleMaintenance.py
```diff
@@ -80,7 +82,8 @@ def register_user():
         if db.session.query(TempUser).filter_by(Username=username).all():
             abort(409, 'User with username {} is pending for permissions'.format(username))
         temp_user = TempUser(Username=username, Password=password, Email=email, ModelsProvider=models_provider,
-                             FirstName=name, LastName=last_name)
+                             FirstName=name, LastName=last_name, Motivation=motivation,
+                             RegistrationDate=datetime.utcnow())
```
SlavomirMazurPantheon, line 86: **RegistrationDatetime** instead of RegistrationDate

api/models.py
```diff
@@ -35,6 +35,8 @@ class BaseUser(DeferredReflection, Base):
     LastName = Column(String(255))
     AccessRightsSdo = Column(String(255))
     AccessRightsVendor = Column(String(255))
+    Motivation = Column(String(255))
```
SlavomirMazurPantheon, line 38: I would use motivation only for users waiting for approval -> users_temp.
After user validation, access rights are set and it is no longer needed to remember motivation.

richardzilincikPantheon:
updated original comment with correct sql

PR #376, opened 2021-10-11
Replace MariaDB with Redis

richardzilincikPantheon:


richardzilincikPantheon:
The following changes are needed on the frontend to make use of this:
- remove `/api/admin/sql-tables` endpoint
- `/api/admin/sql-tables/<table>` needs to be changed to `/api/admin/users/<status>` in all paths
    - valid statuses are `temp` and `approved`
- remove healthcheck for /my-sql

optionally:
- username, first and last name, email and models_provider no longer need to be supplied to the `/api/admin/move-user` endpoint

PR #384, opened 2021-11-15
Create BaseScriptConfig class

richardzilincikPantheon:
Adds a proper class for ScriptConfig classes to inherit from.

PR #386, opened 2021-11-22
statistics.py refactor

richardzilincikPantheon:
depends on #384


statistic/statistics.py
```diff
@@ -106,66 +86,61 @@ def render(tpl_path, context):
...
-            revision = parsed_yang.search('revision')[0].arg
-        except:
-            continue
-        organization = resolve_organization(mod_git, parsed_yang)
-        name = m_name_revision.split('.')[0].split('@')[0]
+        checked[filename] = {}
+        checked[filename]['passed'] = False
+        checked[filename]['in-catalog'] = False
+        revision = None
+        if parsed_yang := yangParser.parse(os.path.abspath(module_path)):
```
SlavomirMazurPantheon, line 129: **TYPO**
 := instead of == 
 in various places in script

SlavomirMazurPantheon: _"parsed_yang"_ is not defined

statistic/statistics.py
```diff
@@ -106,66 +86,61 @@ def render(tpl_path, context):
...
+            if results := parsed_yang.search('revision'):
```
SlavomirMazurPantheon, line 130: _"results"_ is not defined

SlavomirMazurPantheon (changes requested):

statistic/statistics.py
richardzilincikPantheon: := is not a typo, it's a new operator introduced in python 3.8. https://docs.python.org/3/reference/expressions.html#assignment-expressions In each case it only really saves one line of code. If we want to avoid it in our codebase, I can remove it.

SlavomirMazurPantheon: ou sorry, I wasn't familiar with that ... I would probably avoid that for now - we can add this in the future

richardzilincikPantheon: No problem, done.

statistic/statistics.py
```diff
@@ -174,130 +151,118 @@ def get_specifics(path_dir):
...
-    try:
-        table_sdo['percentage_compile'] = repr(round((float(passed) / num_in_catalog) * 100, 2)) + ' %'
-    except ZeroDivisionError:
-        table_sdo['percentage_compile'] = '0.0 %'
-    table_sdo['percentage_extra'] = 'unknown'
-    save_list.append(table_sdo)
-
-
-def solve_platforms(path, platform, LOGGER):
+        modules = int(out.split('{} : '.format(path))[1].splitlines()[0])
```
SlavomirMazurPantheon, line 249:
```
  File "<string>", line 1, in <module>
IndexError: list index out of range
```

statistic/statistics.py
```diff
@@ -569,121 +432,75 @@ def main(scriptConf=None):
...
 
-        for direc in next(os.walk(yang_models + '/vendor'))[1]:
-            vendor_direc = yang_models + '/vendor/' + direc
+        def get_output(**kwargs) -> str:
+            """run runYANGallstats with the provided kwargs as command line arguments.
+            removedup is set to True by default.
+            """
+            kwargs.setdefault('removedup', True)
+            script_conf = all_stats.ScriptConfig()
+            for key, value in kwargs:
```
SlavomirMazurPantheon, line 441: Probably:
```
for key, value in kwargs.items():
```

statistic/statistics.py
```diff
@@ -569,121 +434,75 @@ def main(scriptConf=None):
...
+
+        for direc in next(os.walk(os.path.join(yang_models, 'vendor')))[1]:
+            vendor_direc = os.path.join(yang_models, 'vendor', direc)
             if os.path.isdir(vendor_direc):
-                LOGGER.info('{}'.format(get_curr_dir(__file__) + '/../runYANGallstats/runYANGallstats.py'))
-                process = subprocess.Popen(
-                    ['python', get_curr_dir(__file__) + '/../runYANGallstats/runYANGallstats.py', '--rootdir', vendor_direc,
-                     '--removedup', 'True'], stdout=subprocess.PIPE)
-                out, err = process.communicate()
+                LOGGER.info(os.path.join(get_curr_dir(__file__), 'runYANGallstats.py'))
```
SlavomirMazurPantheon, line 452: This produces following log every time:
```

```

SlavomirMazurPantheon (approved):

PR #393, opened 2021-11-29
Log modules and vendors that failed to patch to confd

richardzilincikPantheon:
Also unifies the code for patching in chunks.

### DIFF
--- a/parseAndPopulate/populate.py
+++ b/parseAndPopulate/populate.py
@@ -241,41 +241,8 @@ def main(scriptConf=None):
     x = -1
     with open('{}/prepare.json'.format(direc)) as data_file:
         read = data_file.read()
-        modules_json = json.loads(read).get('module', [])
-        for x in range(0, int(len(modules_json) / 1000)):
-            json_modules_data = json.dumps({
-                'modules':
-                    {
-                        'module': modules_json[x * 1000: (x * 1000) + 1000]
-                    }
-            })
-
-            if '{"module": []}' not in read:
-                response = confdService.patch_modules(json_modules_data)
-
-                if response.status_code < 200 or response.status_code > 299:
-                    confd_patched = False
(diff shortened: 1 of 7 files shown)


utility/confdService.py
```diff
@@ -43,70 +47,91 @@ def __init__(self):
...
-        return response
-
-    def delete_module(self, module_key: str):
+    def _patch(self, data: list, type: str, log_file: str):
+        errors = False
+        chunk_size = 500
+        chunks = [data[i:i + chunk_size] for i in range(int(len(data)/chunk_size) + 1)]
+        path = '{}/restconf/data/yang-catalog:catalog/{}/'.format(self.confd_prefix, type)
+        for chunk in chunks:
+            self.LOGGER.debug('Sending PATCH request to patch multiple modules')
```
SlavomirMazurPantheon, line 86: Move this log out of for loop.
Add new log in for loop which will l log something like:
```
Processing chunk 2 out of X
```

parseAndPopulate/populate.py
```diff
@@ -237,86 +237,20 @@ def main(scriptConf=None):
...
-
-        if response.status_code < 200 or response.status_code > 299:
-            confd_patched = False
-            path_to_file = '{}/modules-confd-data-rest'.format(direc)
-            with open(path_to_file, 'w') as f:
-                json.dump(json_modules_data, f)
-            LOGGER.error('Request with body {} failed to patch modules with {}'
-                         .format(path_to_file, response.text))
+    modules = json.loads(read).get('module', [])
+    confdService.patch_modules(modules)
```
SlavomirMazurPantheon, line 245: Check whether errors occurred while patching modules too.
```
errors = confdService.patch_modules(modules)
```
Later in script hashes should not be stored if there was a problem with patching modules.

SlavomirMazurPantheon (approved):

PR #395, opened 2021-11-30
Run tests only for python 3.9

richardzilincikPantheon:
Our dockerfile specifies python 3.9, so there should be no need to test on python 3.8. Running tests with 2 versions on github actions also causes race conditions for some tests involving external git repositories, which causes them to fail.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: [3.9]
         redis-version: [6]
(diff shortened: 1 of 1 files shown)


SlavomirMazurPantheon:
The tests were originally intended to run for 3 different versions of Python.  Current (3.9), previous (3.8) and next one. 
However, I think we can change that, so it will only run for the current version - maybe add 3.10 it future before release.

richardzilincikPantheon:
I see, in that case, I think we can set the actions jobs to run one at a time instead of running in parallel.

SlavomirMazurPantheon:
We will address this issue when update to version 3.10 will be needed.

PR #409, opened 2021-12-09
Yang search: sort results by sdo first

### LINKED ISSUE #332: seraching too many modules
When searching for something like "interface" we receive over limited (1000 or 10000) number of entries and it is trimmed to this limmited number.

Three things needs to happen for this to work correctly:

1.  if we have more entries then max amount we need to provide that warn information immediately so the user knows .
2. if we have more entries then max amount we need to give standard organization higher rating so they are shown rather then vendord modules. This comes with problem of search where we actually stop searching after receiving max amount to save time of getting data. Update this to search standard organizations first and then vendors which means we will have to add this information to elastic search
3. the filtering that is done in the output from search in the UI is done only on modules that were received. We need to understand that if we reached maximum amount or entries and user filters for something this filter may apply only on entries that were not searched for at all.

test this well so user don t wait for the output for a long time.

richardzilincikPantheon:
Resolves #332
The `search` endpoint now additionally returns `max-hits` in the response to indicate whether the size limit was hit.


richardzilincikPantheon:
[elasticsearch_sdo_field.txt](https://github.com/YangCatalog/backend/files/7702767/elasticsearch_sdo_field.txt)
I forgot to attach the script to create the fields in elasticsearch required for this sorting. It takes a few minutes for elascicsearch to process the requests (at least on my machine locally).

richardzilincikPantheon:
Replaced by #496

PR #420, opened 2021-12-21
Fix memory leak while parsing modules

richardzilincikPantheon:
Previously, the compilation status data could not
be garbage collected, at it was referenced in each
Modules object and each Modules object was
referenced from the Prepare object. We can let the
garbage collector know we will no longer need this
data after we are done parsing. This allows it to
safely free the memory.

### DIFF
--- a/parseAndPopulate/modules.py
+++ b/parseAndPopulate/modules.py
@@ -220,6 +220,7 @@ class Modules:
             self.__resolve_author_email(author_email)
             self.__resolve_maturity_level(maturity_level)
             self.__resolve_semver()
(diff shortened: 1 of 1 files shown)


richardzilincikPantheon:
I was previously unable to get far when trying to populate modules from from yang/vendors/cisco locally due to the memory usage balooning. I have not yet had time to test on the full cisco directory, but with the fix in place, the growth of memory usage is significantly slower. So far, memory usage for the backend container hasn't gone above 2.5GB.

SlavomirMazurPantheon (approved):

PR #424, opened 2022-01-04
Cleanup and caching 

richardzilincikPantheon:
Caching calls to `utility.yangParser.parse`
makes the first phase of populating modules
several times faster according to my tests.
I don't think this can be done for the
`modulesComplicatedAlgorithms` part without
modifications to pyang.

### DIFF
--- a/api/receiver.py
+++ b/api/receiver.py
@@ -92,11 +92,11 @@ class Receiver:
                 'Failed' or 'Finished successfully'
         """
         all_modules = {}
-        direc = arguments[3]
         tree_created = arguments[-1] == 'True'
         sdo = '--sdo' in arguments
         api = '--api' in arguments
         data_type = 'sdo' if sdo else 'vendor'
+        direc = arguments[3 if sdo else 2]
 
         script_name = 'populate'
         module = __import__('parseAndPopulate', fromlist=[script_name])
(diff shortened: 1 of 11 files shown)


SlavomirMazurPantheon:
I checked what will be logged to the log file if assertion condition is not met. 
We should definitely add some simple error descriptions.

api/receiver.py
```diff
@@ -512,63 +467,65 @@ def run_ietf(self):
...
+            if self.connection:
+                self.connection.close()
         except Exception:
             pass
         self.LOGGER.info('Config loaded succesfully')
 
     def on_request(self, ch, method, properties, body):
         process_reload_cache = multiprocessing.Process(
-            target=self.on_request_thread_safe, args=(ch, method, properties, body,))
+            target=self.on_request_thread_safe, args=(ch, method, properties, body))
```
SlavomirMazurPantheon, line 525:
```
  File "/usr/local/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/local/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
TypeError: on_request_thread_safe() takes 3 positional arguments but 5 were given
```

PR #427, opened 2022-01-13
Speed up docker build

richardzilincikPantheon:
from ~50s to ~10s when a regular file is changed


SlavomirMazurPantheon:
After these changes, there seems to be a problem with running cronjobs - the scripts are not running.
The only script that runs properly is process-changed-mods.py

PR #442, opened 2022-02-25
Add error handling for missing module fields

richardzilincikPantheon:
Specifically, there seem to be modules with the `derived-semantic-version`
field missing. This would then crash running modulesComplicatedAlgorithms
on any new versions of these modules, because their semantic version
couldn't be resolved.


richardzilincikPantheon:
incorporated into #439

Issue #443, opened 2022-03-01
Use HTML in reguler API users email


evyncke:
On the 1st of March, I received an email with "Automatic generated message - RFC IETF" as subject with the list of API users.

The formatting is quite unreadable as it is plain text in a proportional font.

Let's use HTML formatting (with the associated mail header for Content-Type + MIME) for a <table> format.

thanks.

SlavomirMazurPantheon:
Done

PR #445, opened 2022-03-04
Format user reminder emails as HTML

richardzilincikPantheon:


richardzilincikPantheon:
resolves #443

evyncke:
Thank you, it seems that the framework made this easy and convenient.

crontab
```diff
@@ -14,7 +14,7 @@ YANGCATALOG_CONFIG_PATH=<YANGCATALOG_CONFIG_PATH>
 5 5 * * * (cd ~ ; source bin/activate ; echo "`date` starting reviseTreeType" >> /var/yang/logs/crons-log.log ; cd utility ; python reviseTreeType.py)
 5 6 * * * (cd ~ ; source bin/activate; echo "`date` starting openconfigPullLocal" >> /var/yang/logs/crons-log.log ; cd ietfYangDraftPull ; python openconfigPullLocal.py )
 5 12 * * 0 (cd ~ ; source bin/activate ; echo "`date` starting reviseSemver" >> /var/yang/logs/crons-log.log ; cd parseAndPopulate ; python reviseSemver.py)
-5 10 * * * (cd ~ ; source bin/activate ; echo "`date` starting userReminder" >> /var/yang/logs/crons-log.log ; cd utility ; python userReminder.py)
+5 10 * 3,9 1 (cd ~ ; source bin/activate ; echo "`date` starting userReminder" >> /var/yang/logs/crons-log.log ; cd utility ; python userReminder.py)
```
SlavomirMazurPantheon, line 17: Wouldn't it be better if `5 10 * * 1` will be used?

richardzilincikPantheon: The script used to check if it's March or September, but we could have it run every month if that would be preferred. I also made a mistake, the fifth field selects the day of the week, so this would run every Monday, it should be `5 10 1 * *` if we want it to run on the first day of every month.

SlavomirMazurPantheon: I would say that once a week (or once a month) is ideal. Every six months seems like a very long time.

richardzilincikPantheon: Ok, I'll change it to once a month, once a weeks seems excessive given that an email is also sent whenever a new user is registered as well.

PR #450, opened 2022-03-17
Adjust all tests for running locally

richardzilincikPantheon:
After this PR, the only setup that should be needed
to run tests locally is to run prepare_environment.sh,
and feed_redis.py.


richardzilincikPantheon:
resolved #451

PR #468, opened 2022-04-08
Close rabbitmq connection cleanly

richardzilincikPantheon:


api/receiver.py
```diff
@@ -563,6 +557,8 @@ def on_request_thread_safe(self, properties, body_raw: bytes):
                     f.write(new_line)
                 else:
                     f.write(line)
+        if config_reloaded:
+            raise ConfigReloadedException
```
SlavomirMazurPantheon, line 561: Why an exception if we (successfully?) reloaded config?

api/receiver.py
```diff
@@ -581,12 +577,15 @@ def start_receiving(self):
                 self.LOGGER.info('Awaiting RPC request')
 
                 self.channel.start_consuming()
+            except ConfigReloadedException:
+                pass
```
SlavomirMazurPantheon, line 581: Please, at least log what happened, don't just use `pass`.

SlavomirMazurPantheon (changes requested):

api/receiver.py
richardzilincikPantheon: When I was testing, simply calling `stop_consuming()` from within `on_request_thread_safe()` just made the receiver unresponsive without restarting the connection, so I used an exception to break out of `start_consuming()`. But this must have been some error on my end, it seems to work now.

PR #479, opened 2022-04-27
Upgrade flask imports in API scripts

SlavomirMazurPantheon:


richardzilincikPantheon:
Is there any advantage to importing directly from the submodules? All the flask imports are intentionally imported by [flask/__init__.py](https://github.com/pallets/flask/blob/main/src/flask/__init__.py) for public usage.

SlavomirMazurPantheon:
Not sure, I have received a tons of errors from Pylint so I changed it according to Pylint.

richardzilincikPantheon:
Interesting, I wasn't previously using pylint, but it still doesn't complain about the imports on my machine.

SlavomirMazurPantheon:
Sorry I meant Pylance. This is example of an error:
```
"Blueprint" is not exported from module "flask"
Import from "flask.blueprints" instead",
```
Maybe we have different setup for Pylance?
Is it possible to define settings in some file? 
We can push this file directly to the project.

richardzilincikPantheon:
It's possible to have project-level VS Code settings, I agree we should probably do that.

From what I can tell, this error looks like a pyright bug. According to [this](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface):
> Imported symbols are considered private by default. If they use the “import A as A” (a redundant module alias), **“from X import A as A”** (a redundant symbol alias), or “from . import A” forms, symbol “A” is not private unless the name begins with an underscore.

which is exactly what happens in flask's \_\_init\_\_.py file:
```python3
from .blueprints import Blueprint as Blueprint
```
`Blueprint` should be considered public.

SlavomirMazurPantheon:
Project-level VS Code settings sounds good. Can you try to create it when you will have time?

PR #488, opened 2022-05-03
Fetch compilation data more efficiently

richardzilincikPantheon:
Instead of searching over all jsons containing compilation
results, construct the name of the json file we're interested in
and load only that into memory.

It's difficult to measure the decrease in RAM usage properly,
but I have some figures to give an idea:
An idle backend container consumes about 370MB.
With the command
`python populate.py --dir /var/yang/nonietf/yangmodels/yang/vendor/fujitsu/`
old: >2GB in a couple seconds and crashes due to exhausting RAM
new: ~460MB

### DIFF
--- a/parseAndPopulate/groupings.py
+++ b/parseAndPopulate/groupings.py
@@ -27,6 +27,7 @@ import unicodedata
 import xml.etree.ElementTree as ET
 
 from git import InvalidGitRepositoryError
+from git.repo import Repo
 
 import utility.log as log
 from utility import repoutil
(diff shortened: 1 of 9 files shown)


parseAndPopulate/groupings.py
```diff
@@ -96,6 +100,81 @@ def parse_and_load(self):
...
+                version  = split_path[1]
+                return 'NETWORKROUTER{}'.format(alnum(version))
+            elif organization == 'ciena':
+                return 'CIENA'
+            elif organization == 'fujitsu':
+                return 'Fujitsu{}{}'.format(alnum(split_path[1]), alnum(split_path[2]))
+            elif organization == 'nokia':
+                version = split_path[1].removeprefix('latest_sros_')
+                return 'Nokia{}'.format(alnum(version))
+        return None
```
SlavomirMazurPantheon, line 175: Is it not possible to create some kind of nested dictionary that will contain info for each repo? 
It would be stored in staticVaribles.py and this dictionary would be modified instead of a large number of if-else blocks.

richardzilincikPantheon: It would be possible for a few, but for most the process is more complicated than just a search and replace. The only way I see is to make a dictionary with lambda functions, which I think is less clear. The script that creates the files also handles them case by case.

The other option would be to change the way the files are named when they are created. I'd have to check what else depends on the names of the JSONs and HTMLs and it might require wider changes.

SlavomirMazurPantheon (approved):

PR #496, opened 2022-05-09
YANG Search - sort results by SDO first

SlavomirMazurPantheon:


richardzilincikPantheon:
Looks good, doesn't look like anything from #409 is missing.

PR #514, opened 2022-06-03
Run backend and receiver as the yang user

richardzilincikPantheon:
Running parts of the program as root and parts as yang creates
ownership problems and permission errors. Running public facing
applications as root is bad security practise in general.


richardzilincikPantheon:
All tests pass locally (Github actions can't tell us anything because it doesn't use docker images) and there's no complications I can think of, but I wouldn't be surprised if an issue related to this wouldn't show up in tests, because these mostly use different locations for files.
@SlavomirMazurPantheon please have a look if there's anything else you can think of that could cause problems.

SlavomirMazurPantheon:
I can see some problems while running command:
```
docker logs yc-backend
```
Also it seems that cronjobs are not fired by `cron` automatically.

PR #522, opened 2022-06-20
Bugfix: Use parsed yang object with all metadata

SlavomirMazurPantheon:


parseAndPopulate/groupings.py
```diff
@@ -26,10 +26,9 @@
 import unicodedata
 import xml.etree.ElementTree as ET
 
+import utility.log as log
```
richardzilincikPantheon, line 29: Local imports should be grouped separately from third party imports, this should stay as is.

richardzilincikPantheon (changes requested):

parseAndPopulate/groupings.py
SlavomirMazurPantheon: Imports are sorted automatically on save - in the future we can add a setting that will specify how imports should be sorted.

richardzilincikPantheon (approved):

PR #524, opened 2022-06-27
Using SchemaParts dataclass for storing schema parts

SlavomirMazurPantheon:
- parts of the schema are stored in dataclass
and passed as an one argument instead of multiple


richardzilincikPantheon (commented):
Great idea, I think there are other places that could benefit from using dataclasses as well.

PR #539, opened 2022-07-13
chore: handle cases with missing files/directories

SlavomirMazurPantheon:
- handle edge-cases when we are trying to access file/directory
which was not created yet
- removed all yang suite left-over functionality

### DIFF
--- a/api/views/healthCheck/healthCheck.py
+++ b/api/views/healthCheck/healthCheck.py
@@ -399,7 +399,8 @@ def check_cronjobs():
         with open('{}/cronjob.json'.format(ac.d_temp), 'r') as f:
             file_content = json.load(f)
     except (FileNotFoundError, json.decoder.JSONDecodeError):
-        return make_response(jsonify({'error': 'Data about cronjobs are not available.', 'data': {}}), 200)
+        bp.LOGGER.error('cronjob.json file does not exist')
+        file_content = {}
     return make_response(jsonify({'data': file_content}), 200)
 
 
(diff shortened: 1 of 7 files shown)


api/views/healthCheck/healthCheck.py
```diff
@@ -398,8 +398,8 @@ def check_cronjobs():
     try:
         with open('{}/cronjob.json'.format(ac.d_temp), 'r') as f:
             file_content = json.load(f)
-    except Exception:
-        return make_response(jsonify({'error': 'Data about cronjobs are not available.'}), 400)
+    except (FileNotFoundError, json.decoder.JSONDecodeError):
+        return make_response(jsonify({'error': 'Data about cronjobs are not available.', 'data': {}}), 200)
```
richardzilincikPantheon, line 402: Returning an `error` field along with a 200 code doesn't seem right. Maybe we could return a 500 code instead?

richardzilincikPantheon: Or if we consider no data to be a valid response, the error message can be removed entirely.

richardzilincikPantheon (changes requested):

api/views/healthCheck/healthCheck.py
SlavomirMazurPantheon: Yes, you are right - the `error` message in this case is not as important as empty dict in `data` property to correctly render Healthcheck tab in Admin UI

richardzilincikPantheon (approved):

PR #559, opened 2022-08-15
Some performance fix suggestion for parse directory script

Fagtoy:
I spotted that the parse_vendor() function has nested loops, this function loops over all the files in search_directory two times (one time for every pattern), so I rewrote that functionality using only one loop. It's also important that find_files() is a generator, so if the use of the generator and nested loop is made on purpose, this pull request should be closed.

### DIFF
--- a/parseAndPopulate/populate.py
+++ b/parseAndPopulate/populate.py
@@ -88,7 +88,7 @@ class ScriptConfig(BaseScriptConfig):
                 'flag': '--sdo',
                 'help': 'If we are processing sdo or vendor yang modules',
                 'action': 'store_true',
-                'default': False
+                'default': True
             },
             {
                 'flag': '--notify-indexing',
--- a/parseAndPopulate/parse_directory.py
+++ b/parseAndPopulate/parse_directory.py
@@ -26,23 +26,23 @@ __copyright__ = 'Copyright 2018 Cisco and its affiliates, Copyright The IETF Tru
 __license__ = 'Apache License, Version 2.0'
 __email__ = 'miroslav.kovac@pantheon.tech'
 
+import fnmatch
 import glob
-from logging import Logger
 import os
 import shutil
 import time
 import typing as t
+from logging import Logger
 
 import utility.log as log
-from utility.create_config import create_config
-from utility.scriptConfig import Arg, BaseScriptConfig
-from utility.util import find_files, parse_name, parse_revision, strip_comments
-
 from parseAndPopulate.dir_paths import DirPaths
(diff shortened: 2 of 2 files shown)


parseAndPopulate/populate.py
```diff
@@ -88,7 +88,7 @@ def __init__(self):
                 'flag': '--sdo',
                 'help': 'If we are processing sdo or vendor yang modules',
                 'action': 'store_true',
-                'default': False
+                'default': True
```
richardzilincikPantheon, line 91: This should stay as is. With `'action': 'store_true'` and the default being True, there would be no way to specify False.

richardzilincikPantheon (changes requested):
It would be useful to add logging for which xml file is currently being processed.

parseAndPopulate/populate.py
SlavomirMazurPantheon: The main problem is that our default configuration is that `sdo` is False but default `dir` is `/var/yang/nonietf/yangmodels/yang/standard/ietf/RFC`. However, this directory is SDO. If someone runs `populate.py` script with default configuration it results in an error.

Fagtoy:
> It would be useful to add logging for which xml file is currently being processed.

Added logging for which file is being processed

parseAndPopulate/populate.py
richardzilincikPantheon: In that case, we should either remove the `--sdo` flag and replace it with a `--vendor` flag, or add a `--vendor` flag and make passing one of the two required. Otherwise there will be no way to populate vendor directories.

Fagtoy: so can I just change default ```--dir``` to ```/var/yang/nonietf/yangmodels/yang/standard/iana``` and keep default ```--sdo``` as ```False``` then?

SlavomirMazurPantheon: `/var/yang/nonietf/yangmodels/yang/standard/iana` is also SDO. Vendor directories are for example
- /var/yang/nonietf/yangmodels/yang/vendor/cisco
- /var/yang/nonietf/yangmodels/yang/vendor/huawei
- /var/yang/nonietf/yangmodels/yang/vendor/fujitsu

I think you can change `--dir` to the `/var/yang/nonietf/yangmodels/yang/vendor/huawei/network-router/8.20.0/atn980b`

Fagtoy: thanks, done

richardzilincikPantheon (approved):

SlavomirMazurPantheon (approved):

PR #562, opened 2022-08-18
fix f-string in draftPullUtility.py

Fagtoy:


Fagtoy:
fixed f-strings in some scripts, backslash characters are not allowed in f-strings

SlavomirMazurPantheon (approved):

PR #563, opened 2022-08-18
refactor def check_early_revisions() function

Fagtoy:


ietfYangDraftPull/draftPullUtility.py
```diff
@@ -108,52 +108,66 @@ def check_early_revisions(directory: str, LOGGER: logging.Logger) -> None:
         :param directory    (str) full path to directory with yang modules
         :param LOGGER       (logging.Logger) formated logger with the specified name
     """
-    for f in os.listdir(directory):
-        # Extract the YANG module name from the filename
-        module_name = f.split('.yang')[0].split('@')[0]   # Beware of some invalid file names such as '@2015-03-09.yang'
+    get_module_name = lambda filename: filename.split('.yang')[0].split('@')[0]
```
richardzilincikPantheon, line 111: A full function definition would be more easily visible than assigning lambdas to variables. Also makes it easier to add type hints.

richardzilincikPantheon (changes requested):

ietfYangDraftPull/draftPullUtility.py
Fagtoy: replaced lambda with ordinary function

richardzilincikPantheon: same for `is_revision_part_valid`

Fagtoy: fixed

richardzilincikPantheon (approved):

ietfYangDraftPull/draftPullUtility.py
```diff
@@ -108,52 +108,72 @@ def check_early_revisions(directory: str, LOGGER: logging.Logger) -> None:
...
-            os.remove(os.path.join(directory, fi))
+            if os.path.exists((path := os.path.join(directory, file_to_delete))):
+                os.remove(path)
+                
+                
+def get_module_name(filename: str) -> str:
+    return filename.split('.yang')[0].split('@')[0]
+
+
+def is_revision_part_valid(revision_part: str) -> str:
```
SlavomirMazurPantheon, line 153: This method is returning `bool`

ietfYangDraftPull/draftPullUtility.py
```diff
@@ -108,52 +108,72 @@ def check_early_revisions(directory: str, LOGGER: logging.Logger) -> None:
...
-            if 'iana-if-type' in fi:
+        for file_to_delete in files_to_delete:
+            if 'iana-if-type' in file_to_delete:
                 break
-            os.remove(os.path.join(directory, fi))
+            if os.path.exists((path := os.path.join(directory, file_to_delete))):
+                os.remove(path)
+                
+                
+def get_module_name(filename: str) -> str:
```
SlavomirMazurPantheon, line 149: Please, put underscore before these helper methods which are called from `check_early_revisions()` to make them only internal.
Also maybe put them at the end of the file

ietfYangDraftPull/draftPullUtility.py
Fagtoy: fixed

ietfYangDraftPull/draftPullUtility.py
Fagtoy: I put underscores before these functions, but I don't think moving these functions at the end of the file is better, because when you check what a function does then you need to go back 100+ lines of code to find where it is used, it's not convenient

PR #566, opened 2022-08-23
refactored code and added In Progress job_log for scripts

Fagtoy:
tried to minimize intends
replaced empty dictionaries creating, and then assigning key-value pairs to them with creating dictionaries with already filled key/values
replaced .format() with f-strings
added job_log() function with "In Progress" status to scripts that use job_log() function with "Failure"/"Success" statuses

### DIFF
--- a/api/matomo_tracker.py
+++ b/api/matomo_tracker.py
@@ -30,9 +30,7 @@ def get_headers_dict(req) -> dict:
         'PATH_INFO',
         'QUERY_STRING',
     ]
-    data = {
-        'HTTPS': req.is_secure
-    }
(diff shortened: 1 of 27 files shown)


SlavomirMazurPantheon:
Refactoring parts looks good, I like these changes.

SlavomirMazurPantheon:
What about using `Enum` for the status of the job?

Fagtoy:
> What about using `Enum` for the status of the job?

Yeah, it would be great, I'll do it

Fagtoy:
> What about using `Enum` for the status of the job?

done

SlavomirMazurPantheon (approved):

PR #580, opened 2022-08-29
feat: added a monthly job "redis_users_recovery.py"

### LINKED ISSUE #570: Run redisUsersRecovery.py script once a month
Run `redisUsersRecovery.py` once a month as a cronjob (see crontab files)
- Check whether script is still working (some users needs to be created - using Admin UI for example)
- Check whether saving and loading is working
- Rename script to follow snake_case: `redis_users_recovery.py` (**NOTE**: Then it needs to be renamed in all imports too!)
- This script uses `redisUsersConnection.py` - also apply snake_case and move it to the `redisConnections` directory
- Other refactoring changes may be applied too

Fagtoy:
renamed redisUsersRecovery.py and redisUsersConnection.py in snake case, moved redis_users_connection.py to the redisConnections folder. I moved logic in redis_users_recovery.py to a class with methods for saving/loading data to/from backups, it seems more convenient now. done some refactoring in admin views (mostly replaced .format() with f-strings or os.path.join()). added a monthly cronjob for saving data from Redis db to a backup.

resolves #570


richardzilincikPantheon (changes requested):

recovery/redis_users_recovery.py
```diff
@@ -0,0 +1,163 @@
...
+                elif key_type == 'set':
+                    value = [member.decode() for member in self.redis.smembers(key)]
+                elif key_type == 'hash':
+                    hash_table = self.redis.hgetall(key)
+                    value = {hash_key.decode(): hash_table[hash_key].decode() for hash_key in hash_table}
+                else:
+                    exception = ValueError(
+                        f'Key of unknown type ({key_type}) was found while saving data from redis'
+                    )
+                    self.log_and_raise_exception(exception)
```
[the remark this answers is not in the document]
Fagtoy: done

recovery/redis_users_recovery.py
```diff
@@ -102,18 +102,26 @@ def backup_data_from_redis(self):
...
+                    elif key_type == 'hash':
+                        hash_table = self.redis.hgetall(key)
+                        value = {hash_key.decode(): hash_table[hash_key].decode() for hash_key in hash_table}
+                    else:
+                        raise ValueError(
+                            f'Key of unknown type ({key_type}) was found while saving data from redis'
+                        )
+                except ValueError as e:
+                    exception_message = str(e)
+                    self.logger.exception(exception_message)
```
richardzilincikPantheon, line 119: It's not necessary to pass the exception message as an argument here. `logger.exception()` fetches the message from the currently active exception, so it would appear in the log output twice.

richardzilincikPantheon (changes requested):

recovery/redis_users_recovery.py
Fagtoy: Fixed

richardzilincikPantheon (approved):

SlavomirMazurPantheon (approved):

PR #581, opened 2022-08-30
fix: handle platform names with slashes

richardzilincikPantheon:


richardzilincikPantheon:
The `sandbox/reencode_redis_keys.py` script needs to be run on the redis database before this is deployed. The best way to do this might be to download the prod database locally, run this script on it, and then upload it back to prod before the new release is brought up.

SlavomirMazurPantheon (approved):

PR #582, opened 2022-08-30
fix: added type of exception catched by except statement

### LINKED ISSUE #576: Always specify exception type
In try-except, exception type should be always specified.
From the [PEP-8](http://www.python.org/dev/peps/pep-0008/) Style Guide for Python :
`When catching exceptions, mention specific exceptions whenever possible instead of using a bare except: clause.`
For example, use:
```
try:
  print(x)
except NameError:
  print('Variable x is not defined')
```
instead of bare `exception:`
```
try:
  print(x)
except:
  print('An exception occurred')
```
This task should be completed in two steps/phases:
- [X] Replace `exception:` at least with `exception Exception:`
- [ ] Define specific exceptions - if not sure which type of exception to use try to think whether exception is necessary of think about using `finally`

It would also be beneficial to use logger.exception() in Exception block (if logger available) to log the exception properly.

Fagtoy:
specified the type of exceptions caught by ```except:``` statement, added at least ```except Exception:``` where the type of exception is not clear, removed ```try/except``` blocks where they are not needed

fixes #576


parseAndPopulate/groupings.py
```diff
@@ -490,14 +490,10 @@ def parse_and_load(self):
                 self._parse_raw_capability(module.text)
             modules = self.root.iter('{}capability'.format(tag.split('hello')[0]))
 
-        try:
-            schema_parts = SchemaParts(
-                repo_owner=self.repo_owner, repo_name=self.repo_name,
-                commit_hash=self.commit_hash, submodule_name=self.submodule_name)
-        except:
-            LOGGER.exception('Missing attribute, likely caused by a broken path in {}/platform-metadata.json'
```
richardzilincikPantheon, line 498: Are you sure this can be safely removed? If [this](https://github.com/YangCatalog/backend/blob/80c4b2b328f47b725179545132674af40c4d2fda/parseAndPopulate/groupings.py#L389) check doesn't pass, I think the attributes will remain uninitialized.

parseAndPopulate/integrity.py
```diff
@@ -86,27 +86,26 @@ def __init__(self):
...
         if revision_parts[1:] == [2, 29]:
             revision_parts[2] = 28
             try:
                 date(*revision_parts)
-            except:
+            except ValueError:
                 return False
-        else:
-            return False
+        return False
```
richardzilincikPantheon, line 101: This would return False, even if we can fix the date.

parseAndPopulate/modulesComplicatedAlgorithms.py
```diff
@@ -633,15 +633,16 @@ def add_to_new_modules(new_module: dict):
...
-                                    new_yang_tree, old_yang_tree = trees
-                                    if trees_match(new_yang_tree, old_yang_tree):
-                                        # yang trees are the same - update only the patch version
-                                        update_semver(newest_existing_module_details, new_module, PATCH)
-                                    else:
-                                        # yang trees have changed - update minor version
-                                        update_semver(newest_existing_module_details, new_module, MINOR)
-                            except:
+                                if not trees:
+                                    pass
```
richardzilincikPantheon, line 637: `continue`?

parseAndPopulate/modulesComplicatedAlgorithms.py
```diff
@@ -670,17 +671,18 @@ def add_to_new_modules(new_module: dict):
...
-                                            # yang trees are the same - update only the patch version
-                                            update_semver(prev_module_details, module, 2)
-                                            curr_module_details.semver = increment_semver(prev_module_details.semver, 2)
-                                        else:
-                                            # yang trees have changed - update minor version
-                                            update_semver(prev_module_details, module, 1)
-                                            curr_module_details.semver = increment_semver(prev_module_details.semver, 1)
-                                except:
+                                    if not trees:
+                                        pass
```
richardzilincikPantheon, line 675: same as above

richardzilincikPantheon (changes requested):
Looks like `TestApiInternalClass.test_get_local_commit_file_not_found` is failling in this branch, while it's passing in master.

api/sender.py
```diff
@@ -96,32 +97,30 @@ def send(self, arguments: list) -> str:
...
-                              )
-        with open('{}/{}'.format(self._temp_dir, self._response_file), 'a') as f:
-            line = '{} -- {} - {}\n'.format(datetime.datetime.now().ctime(),
-                                            corr_id, StatusMessage.IN_PROGRESS.value)
+        channel.basic_publish(
+            exchange='', routing_key='module_queue', body=str(arguments),
+            properties=pika.BasicProperties(correlation_id=corr_id),
+        )
+        with open(os.path.join(self._temp_dir, self._response_file), 'a') as f:
+            line = f'{datetime.datetime.now().ctime()} -- {corr_id} - {StatusMessage.IN_PROGRESS}\n'
```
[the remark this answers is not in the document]
Fagtoy: there's no need for that, I checked it and enums with f-strings work fine that way
image

parseAndPopulate/integrity.py
Fagtoy: I just removed the ```else``` statement, that doesn't really change anything, so ```if revision_parts[1:] != [2, 29]:``` then ``False``` will be returned, we just can skip the ```else``` statement here

api/sender.py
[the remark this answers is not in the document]
richardzilincikPantheon: Oh ok. You'll have to make `StatusMessage` inherit from `str` then, it doesn't seem to work without that.

parseAndPopulate/groupings.py
Fagtoy: Yeah, I didn't see it, fixed, thanks

api/sender.py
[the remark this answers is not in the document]
Fagtoy: done

parseAndPopulate/integrity.py
richardzilincikPantheon: Yes, but now `False` will also be returned even if `revision_parts[1:] == [2, 29]` is `True`.

richardzilincikPantheon: Yes, but now `False` will also be returned even if `revision_parts[1:] == [2, 29]` is `True`.

Fagtoy: Oh, I see, thanks, fixed

Fagtoy:
> Looks like `TestApiInternalClass.test_get_local_commit_file_not_found` is failling in this branch, while it's passing in master.

fixed

richardzilincikPantheon (approved):

SlavomirMazurPantheon (approved):

PR #585, opened 2022-09-01
fix: added logging for which file is being deleted at the moment from es

### LINKED ISSUE #573: Add log which module was removed from Elasticsearch
As in the title, name@revision is not logged when deleting modules from Elasticsearch.
To test deleting modules from Elasticsearch use `yang2_repo_deletes.dat` file.
Content of the file should look like:
```
[
module_name@revision/organization,
another_module_name@revision/organization
...
]
```

Fagtoy:
refactored process-changed-mods.py script by using a separate class for processing changed mods, added logging for which module is currently deleting from es
couldn't check if it's working or not because my ```yang2_repo_deletes.dat``` and ```yang2_repo_cache.dat``` files are empty

resolves #573


elasticsearchIndexing/process-changed-mods.py
```diff
@@ -23,186 +23,223 @@
...
+        self.delete_cache = self._load_delete_cache(self.delete_cache_path)
+        
+        if not self.changes_cache and not self.delete_cache:
+            self.logger.info('No new modules are added or removed. Exiting script!!!')
+            os.unlink(self.lock_file)
+            os.unlink(self.lock_file_cron)
+            sys.exit()
+
+        self.logger.info('Pulling latest YangModels/yang repository')
+        repoutil.pull(self.yang_models)
```
Fagtoy, line 92: I also have a question about pulling yang_models every time the script executes, what's the point of pulling models? I don't see where these models are used then, maybe I'm missing something.

richardzilincikPantheon: Hmm, I'm not 100% sure either. Looks like it was introduced in [this commit](https://github.com/YangCatalog/backend/commit/7236d4be063fcca99a230e6d1cddf3fca70a0c96#diff-edbe1631ff443625e1373391fe15a0cde434a4905188667d30d94975b9a448aa). @SlavomirMazurPantheon do you know if this is important for something?

richardzilincikPantheon:
> couldn't check if it's working or not because my `yang2_repo_deletes.dat` and `yang2_repo_cache.dat` files are empty

Just editing `yang2_repo_deletes.dat` and adding something like
```json
[
"ietf-alarms@2019-09-11/ietf",
]
```
should work I think.

Fagtoy:
> 

thanks, it seems everything works fine, there are
```Deleting ietf-alarms@2019-09-11/ietf from es indices```
log in ```process-changed-mods.log```
and
```
es_manager.py es-manager => Deleting module: "{'name': 'ietf-alarms', 'revision': '2019-09-11', 'organization': 'ietf'}" from index: "ESIndices.YINDEX" - 160
es_manager.py es-manager => Deleting module: "{'name': 'ietf-alarms', 'revision': '2019-09-11', 'organization': 'ietf'}" from index: "ESIndices.MODULES" - 160
es_manager.py es-manager => Deleting module: "{'name': 'ietf-alarms', 'revision': '2019-09-11', 'organization': 'ietf'}" from index: "ESIndices.AUTOCOMPLETE" - 160
es_manager.py es-manager => Deleting module: "{'name': 'ietf-alarms', 'revision': '2019-09-11', 'organization': 'ietf'}" from index: "ESIndices.TEST" - 160
```
in ```es-manager.log```

elasticsearchIndexing/process-changed-mods.py
Fagtoy: @SlavomirMazurPantheon it takes a lot of time to finish as well, so if it's not important I'll better remove it

SlavomirMazurPantheon: I am not sure why I put this here - it can be removed, because this script runs every 3 minutes while the yangmodels repo is updated 1-2 a day so it is not necessary to check changes every 3 minutes.
In fact, maybe this was the reason, that I wanted this local cloned repo to be as actual as possible.

However, question is why it is taking so much time when it should perform just `git pull`?

Fagtoy: I just wanted to check how much time it takes to make a pull request here but couldn't do it because I have a broken ```config``` file in ```/var/yang/nonietf/yangmodels/yang/.git```, so @richardzilincikPantheon @SlavomirMazurPantheon could anyone please share his ```config``` file with me and I will check it then?
I didn't notice an error earlier, so I believe this part doesn't really consume so much time because I just commented it earlier to check if logging works, as I saw that the script wasn't passing this ```pull``` request for a long time

richardzilincikPantheon (approved):

PR #586, opened 2022-09-01
fix: removed outdated mentions of ConfD

### LINKED ISSUE #574: Remove outdated mentions of ConfD
ConfD was originally used as the primary database for yangcatalog, but it has since been replaced by Redis, and is now only used for verifying data integrity. Many mentions in comments, variable and function names are still present in the codebase. There are also several unused functions for communicating with ConfD.
Comments, variable and function names need to be updated to avoid confusion and reflect the current status of the code. Unused code should be removed.

Fagtoy:
removed outdated mentions of ConfD and ConfD functions that are no longer used

fixes #574


richardzilincikPantheon:
I think the `ConfdService` in `receiver.py` can also be removed, doesn't look like it's accessed anywhere. Same for the confd mocks in `test_receiver.py`.

Fagtoy:
> I think the `ConfdService` in `receiver.py` can also be removed, doesn't look like it's accessed anywhere. Same for the confd mocks in `test_receiver.py`.

done

richardzilincikPantheon (approved):

SlavomirMazurPantheon (approved):

PR #588, opened 2022-09-02
feat: added mutually exclusive arguments to BaseScriptConfig

### LINKED ISSUE #587: Add a way to specify mutually exclusive arguments to scripts that use ScriptConf
Scripts that can be run from the admin UI have their command line options specified by a `ScriptConfig` class. There is currently no way to tell `ScriptConfig` to make a group of arguments mutually exclusive.
This would be particularly useful for `recovery.py` and `elk_recovery.py`. `recovery.py` currently uses a `--type` argument along with `--name_load` and `--name_save`. `elk_recovery.py` allows specifying both a `--save` argument, and a `--load` argument that overrides it. Neither of these approaches is ideal.

The information about arguments being mutually exclusive should also be made available on the `/api/admin/scripts/<script>` endpoint to allow making it visible on the frontend in some way.

### LINKED ISSUE #590: add succsessful job_log message on ```recovery.py``` script finish
now there is no success message for ```recovery.py``` script with ```load``` type

Fagtoy:
added mutually_exclusive arguments to BaseScriptConfig, added information whether a particular argument is mutually exclusive with any other arguments into ```args_dict``` attribute, for example, information of ```api/admin/scripts/elk_recovery``` looks like this now
image
and response for ```api/admin/scripts/recovery``` looks like this
image

resolves #587 
resolves #590


utility/scriptConfig.py
```diff
@@ -45,22 +47,60 @@ class BaseScriptConfig:
...
+                
+    def _add_args(self, args: list[Arg]):
+        for arg in args:
+            flag = arg.pop('flag')
+            if flag in self.mutually_exclusive_arg_flags:
+                continue
+            self.parser.add_argument(flag, **arg)
+            self._add_argument_to_args_dict(flag, arg)
+        
+    def _add_argument_to_args_dict(self, flag: str, arg: Arg):
```
richardzilincikPantheon, line 94: It might be better to accept the `arg_name` instead of the flag, the same way `_update_argument_in_args_dict()` does. You should be able to get the `.dest` from the return value of `parser.add_argument()`.

richardzilincikPantheon (changes requested):
Now that we can use mutually exclusive arguments, I think it would be better to have the flags work the same way in all the recovery scripts. `--save` and `--load` flags to specify the type of operation, and a `--file` flag that would specify the filename for either operation.
`redisUsersRecovery.py` should also be updated.

recovery/elk_recovery.py
```diff
@@ -35,7 +35,23 @@ class ScriptConfig(BaseScriptConfig):
 
     def __init__(self):
         help = __doc__
-        args: t.List[Arg] = [
+        mutually_exclusive_args: tuple[tuple[Arg]] = (
```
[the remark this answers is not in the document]
Fagtoy: I changed ```tuple``` to ```list```, python allows using ```list``` in type hints, why use ```t.List``` then?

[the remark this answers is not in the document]
richardzilincikPantheon: Oh, you're right. Looks like this was introduced in python 3.9 and I didn't notice. Just `list` is fine then.

Fagtoy:
> Now that we can use mutually exclusive arguments, I think it would be better to have the flags work the same way in all the recovery scripts. `--save` and `--load` flags to specify the type of operation, and a `--file` flag that would specify the filename for either operation. `redisUsersRecovery.py` should also be updated.

done

recovery/elk_recovery.py
```diff
@@ -89,18 +87,18 @@ def main(scriptConf=None):
...
     else:
-        if not args.latest:
-            snapshot_name = args.name_load
-
-        sorted_snapshots = es_snapshots_manager.get_sorted_snapshots()
-        if not sorted_snapshots:
-            print('There are no snapshots to restore')
-            sys.exit(1)
-        snapshot_name = sorted_snapshots[-1]['snapshot']
+        if args.latest:
```
richardzilincikPantheon, line 92: Shouldn't this stay as `if not args.latest:`?

recovery/recovery.py
```diff
@@ -146,10 +150,9 @@ def main(scriptConf=None):
...
         LOGGER.info('Save completed successfully')
         job_log(start_time, temp_dir, messages=messages, status=JobLogStatuses.SUCCESS, filename=current_file_basename)
-    else:
-        file_name = ''
-        if args.name_load:
-            file_name = os.path.join(confd_backups, args.name_load)
+    elif args.load:
+        if args.file:
+            file_name = os.path.join(confd_backups, args.file)
         else:
```
richardzilincikPantheon, line 156: `args.file` will never be `None` since it has a default value.

recovery/redis_users_recovery.py
```diff
@@ -111,29 +115,27 @@ def backup_data_from_redis(self):
...
-        self.logger.info(f'Data saved to {self.args.name_save} successfully')
-
+        self.logger.info(f'Data saved to {self.args.file} successfully')
 
     def load_data_from_backup_to_redis(self):
-        if self.args.name_load:
-            file_name = f'{os.path.join(self.backups, self.args.name_load)}.json'
+        if self.args.file:
+            file_name = f'{os.path.join(self.backups, self.args.file)}.json'
         else:
```
richardzilincikPantheon, line 139: Same as above.

richardzilincikPantheon (changes requested):
The current date time isn't really a good default for the load operation as it will never exist. Maybe we could just leave the default value of `args.file` empty, and fill it in with either the date time, or the latest backup, depending on the type of operation.

recovery/recovery.py
Fagtoy: I assume that we can provide an empty value if we use a load process, as the default process is saving and we provide a default value to ```file``` for this

Fagtoy:
> The current date time isn't really a good default for the load operation as it will never exist. Maybe we could just leave the default value of `args.file` empty, and fill it in with either the date time, or the latest backup, depending on the type of operation.

Okay, this approach seems more convinient to me, I'll try to pass correct datetime as an argument to file for the ```redis_users_recovery``` cronjob

recovery/recovery.py
richardzilincikPantheon: It's not actually possible to start the load process. Even if the `--load` flag is passed, `args.save` is checked first, and it's default value is `True`. You can try running the script and checking `/var/yang/logs/yang.log`. The default value for `args.file` is set regardless of the `--load` and `--save` flags.

Fagtoy: > The default value for args.file is set regardless of the --load and --save flags.

Yeah, I understand that.

> It's not actually possible to start the load process. Even if the --load flag is passed, args.save is checked first, and it's default value is True

Maybe we should set the default values for both ```save``` and ```load``` to ```False``` then and specify the particular flag for each run of the script explicitly?

richardzilincikPantheon: Yes, that might be better.

richardzilincikPantheon:
Looks good, one last thing, the descriptions for `--file` should be updated to say what the default value is for load operations as well, and the `--latest` flag can be removed from `elk_recovery.py` so it behaves the same way as the other two scripts.

richardzilincikPantheon:
`recovery.py` also needs the `--load` flag in crontab

Fagtoy:
> Looks good, one last thing, the descriptions for `--file` should be updated to say what the default value is for load operations as well, and the `--latest` flag can be removed from `elk_recovery.py` so it behaves the same way as the other two scripts.

done

> recovery.py also needs the --load flag in crontab

there is no crontab job for this, this script is run by the start of the api-recovery container

@SlavomirMazurPantheon please take a look at this PR, I also included fixes for [this issue](https://github.com/YangCatalog/backend/issues/590) here, and when this PR is merged please merge [this PR](https://github.com/YangCatalog/deployment/pull/153) as well to have a proper command in the api-recovery container for starting ```recovery.py``` script

richardzilincikPantheon:
> there is no crontab job for this, this script is run by the start of the api-recovery container

https://github.com/YangCatalog/backend/blob/master/crontab#L22
It's a save not a load though.

Fagtoy:
> https://github.com/YangCatalog/backend/blob/master/crontab#L22 It's a save not a load though.

I didn't see this one, thanks, fixed

richardzilincikPantheon (approved):

recovery/elk_recovery.py
```diff
@@ -35,45 +35,41 @@ class ScriptConfig(BaseScriptConfig):
...
-                'flag': '--name_save',
-                'help': 'Set name of the file to save. Default name is date and time in UTC',
-                'type': str,
-                'default': datetime.datetime.utcnow().strftime(backup_date_format).lower()
-            },
+        mutually_exclusive_args: list[list[Arg]] = [
+            [
+                {
+                    'flag': '--save',
+                    'help': 'Set whether you want to create snapshot. Default is True',
```
SlavomirMazurPantheon, line 42: `Default is True` in help, but default set to False

recovery/recovery.py
```diff
@@ -188,6 +195,10 @@ def main(scriptConf=None):
             redisConnection.reload_modules_cache()
             redisConnection.reload_vendors_cache()
             LOGGER.info('All the modules data set to Redis successfully')
+            job_log_messages.extend([
+                {'label': 'Loaded modules', 'message': len(modules)},
+                {'label': 'Loaded vendors', 'message': len(vendors)}
+            ])
```
SlavomirMazurPantheon, line 201: It is not really important to job_log how many modules was loaded. However, if we want to log this information we need to distinguish between `load` and `save` -> see current dashboard in admin ui, there is `recovery - save`

SlavomirMazurPantheon (approved):

PR #592, opened 2022-09-07
refactor: recovery.py script

### LINKED ISSUE #591: Refactor recovery.py script
I noticed that ```recovery.py``` script is big and difficult to understand, so I think it would be better to refactore it as ```redis_users_recovery.py```, to decouple different methods, and to make the script more readable

Fagtoy:
refactored recovery.py script using separate classes with decoupled methods for each type of script (```--save```, ```--load```)

resolves #591

### DIFF
--- a/recovery/recovery.py
+++ b/recovery/recovery.py
@@ -69,8 +69,8 @@ class ScriptConfig(BaseScriptConfig):
             {
                 'flag': '--file',
                 'help': (
-                    'Set name of the file to save data to/load data from. Default name is empty. '
-                    'If name is empty: load operation will use the last available backup file (rdb or json), '
+                    'Set name of the file (without file format) to save data to/load data from. Default name is empty. '
+                    'If name is empty: load operation will use the last available json backup file, '
                     'save operation will use date and time in UTC.'
                 ),
                 'type': str,
@@ -188,7 +188,7 @@ class LoadDataFromBackupToDatabase(Recovery):
 
     def _start_process(self):
         if self.args.file:
-            self.args.file = os.path.join(self.redis_json_backup, self.args.file)
+            self.args.file = os.path.join(self.redis_json_backup, f'{self.args.file}.json')
         else:
             list_of_backups = get_list_of_backups(self.redis_json_backup)
             self.args.file = os.path.join(self.redis_json_backup, list_of_backups[-1])
@@ -214,13 +214,7 @@ class LoadDataFromBackupToDatabase(Recovery):
     def _load_data_from_redis_backup(self) -> tuple[list, list]:
         modules = []
         vendors = []
-        if self.args.file.endswith('.gz'):
-            with gzip.open(self.args.file, 'r') as file_load:
-                self.logger.info(f'Loading file {file_load.name}')
-                catalog_data = json.loads(file_load.read().decode())
-                modules = catalog_data.get('yang-catalog:catalog', {}).get('modules', {}).get('module', [])
-                vendors = catalog_data.get('yang-catalog:catalog', {}).get('vendors', {}).get('vendor', [])
-        elif self.args.file.endswith('.json'):
+        if self.args.file.endswith('.json'):
             with open(self.args.file, 'r') as file_load:
                 self.logger.info(f'Loading file {file_load.name}')
                 catalog_data = json.load(file_load)
--- a/recovery/recovery.py
+++ b/recovery/recovery.py
@@ -32,6 +32,7 @@ import json
 import os
 import time
 import typing as t
+from configparser import ConfigParser
 from time import sleep
 
 import utility.log as log
@@ -50,7 +51,6 @@ class ScriptConfig(BaseScriptConfig):
 
     def __init__(self):
         help = __doc__
-        config = create_config()
         mutually_exclusive_args: list[list[Arg]] = [
             [
                 {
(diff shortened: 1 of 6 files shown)


richardzilincikPantheon (commented):
I think this script could actually benefit from several changes. Looks like it only lets you specify the file name for the confd backup, which we don't really care about much anymore, and not for the redis backup. 
@SlavomirMazurPantheon Do we still have any need to back up confd at all since we never retrieve data from it anyway? I think we might be able to remove that entirely.
If we can, it would probably be better to use the file name passed with `--file` for the redis backup (instead of overriding it with `backup.json`).

SlavomirMazurPantheon:
We don't need to backup ConfD at all currently, as it not contains all the modules all the time.
It would be better to backup Redis as you mentioned. These should be .rdb file backup and also some json file which can be used for possible re-populating of the Redis database from file.
NOTE: Don't forget to update removeUnused.py script if you will change something - so the created backup are also removed after some period correctly.

Fagtoy:
Removed loading of backup data from ConfD, and added ```--rdb_file``` argument for saving data of Redis ```rdb``` file. There is also a possible issue with rdb loading data from the ```rdb``` backup because if ```--file``` argument is not provided to the ```load``` operation, we will always backup data from the lates ```json``` backup and not the latest ```rdb``` backup as the ```json``` backup file is the latest in backups directory

richardzilincikPantheon:
> There is also a possible issue with rdb loading data from the `rdb` backup because if `--file` argument is not provided to the `load` operation, we will always backup data from the lates `json` backup and not the latest `rdb` backup as the `json` backup file is the latest in backups directory

Yeah this is intentional. It's not a good idea to overwrite redis' rdb file while it's up and running, so redis needs to be stopped when we want to restore an rdb file. The rdb file also contains the user database and some other things that would also get overwritten.
So it's not really possible to restore the rdb from this script, that's something that would have to be done manually.

Fagtoy:
> > There is also a possible issue with rdb loading data from the `rdb` backup because if `--file` argument is not provided to the `load` operation, we will always backup data from the lates `json` backup and not the latest `rdb` backup as the `json` backup file is the latest in backups directory
> 
> Yeah this is intentional. It's not a good idea to overwrite redis' rdb file while it's up and running, so redis needs to be stopped when we want to restore an rdb file. The rdb file also contains the user database and some other things that would also get overwritten. So it's not really possible to restore the rdb from this script, that's something that would have to be done manually.

So should I remove this functionality from this script for now?

richardzilincikPantheon:
I'd leave the functionality for createing rdb backups just in case something unexpected happens and we'll want to manually restore.

recovery/recovery.py
```diff
@@ -74,18 +70,32 @@ def __init__(self):
                 'flag': '--file',
                 'help': (
                     'Set name of the file to save data to/load data from. Default name is empty. '
-                    'If name is empty: load operation will use the last backup file, '
+                    'If name is empty: load operation will use the last available backup file (rdb or json), '
```
richardzilincikPantheon, line 73: The script only loads json files.

richardzilincikPantheon (changes requested):

Fagtoy:
> I'd leave the functionality for createing rdb backups just in case something unexpected happens and we'll want to manually restore.

Sorry, I wasn’t clear. I’m talking about loading data from an rdb backup, just to be sure that there will be no problems in the future if an rdb backup is loaded later than a json backup (for example, due to some wrong refactoring of the script by future us or other developers)

recovery/recovery.py
Fagtoy: Not really, we can specify an rdb backup file as an argument https://github.com/YangCatalog/backend/blob/f4333d52155752cd2915fea724526f85ad5d089b/recovery/recovery.py#L177, so that’s why I’ll better remove this functionality entirely for now

richardzilincikPantheon:
rdb backups are stored in a separate directory. When looking for the latest backup, the script only looks for files in the directory with jsons, so it should be fine.

recovery/recovery.py
richardzilincikPantheon: Ooh I see. Well this would work for gzipped json files, but we don't create those anywhere, so it would be less confusing to remove it, yeah.

Fagtoy:
I also faced and fixed an edge case. When there's no backup file created yet, so the load process would not run properly, and there will be a hanging "In Progress" recovery cronjob at healthcheck endpoint, so now I log this information and exit the script right there

utility/create_config.py
```diff
@@ -4,6 +4,6 @@
 
 def create_config(config_path=os.environ['YANGCATALOG_CONFIG_PATH']):
     config = configparser.ConfigParser()
-    config._interpolation = configparser.ExtendedInterpolation() # pyright: ignore
+    config._interpolation = configparser.ExtendedInterpolation()  # pyright: ignore
```
SlavomirMazurPantheon, line 7: Please change this to: 
`config = configparser.ConfigParser(interpolation=configparser.ExtendedInterpolation())`
also change it for all the occurrences in this `backed` repo

recovery/redis_users_recovery.py
```diff
@@ -82,11 +82,11 @@ def __init__(self, script_conf: BaseScriptConfig = ScriptConfig(), config: Confi
         self.start_time = None
         self.args = script_conf.args
         self.config = config
```
SlavomirMazurPantheon, line 84: `config` is not used anywhere outside this `__init__` method so it don't need to be in `self`

recovery/recovery.py
```diff
@@ -71,73 +70,104 @@ def __init__(self):
...
+            redis_connection: RedisConnection = RedisConnection()
+    ):
+        self.start_time = None
+        self.job_log_messages = []
+        self.job_log_filename = current_file_basename
+        self.yang_catalog_module_name = 'yang-catalog@2018-04-03/ietf'
+        self.process_type = ''
+
+        self.args = args
+        self.config = config
```
SlavomirMazurPantheon, line 107: `config` is not used anywhere outside this `__init__` method so it don't need to be in `self`

PR #605, opened 2022-09-14
docs: add an overview section to the README

richardzilincikPantheon:


SlavomirMazurPantheon:
Thanks for cronjobs list update, but wouldn't it be better if it was like:
[ianaPull.py](https://github.com/YangCatalog/backend/blob/master/ietfYangDraftPull/ianaPull) - rsyncs and populates new modules from IANA

richardzilincikPantheon:
Yeah that makes more sense, I'll update it.

SlavomirMazurPantheon (approved):

PR #606, opened 2022-09-14
added ImplementationResolver and changed tests for VendorModule

dmytro-kyrychenko:


SlavomirMazurPantheon (approved):

SlavomirMazurPantheon:
Just one last detail - please rename `implementation.py` to `implementations.py` so it matches name of the property that it resolves.

PR #610, opened 2022-09-15
fix: fixed path to the file while storing hashes

SlavomirMazurPantheon:
Signed-off-by: SlavomirMazurPantheon <slavomir.mazur@pantheon.tech>


richardzilincikPantheon:
More comprehensive solution in #616

PR #612, opened 2022-09-16
Break down resolveExpiration.py script

dmytro-kyrychenko:


parseAndPopulate/resolvers/expiration.py
```diff
@@ -0,0 +1,124 @@
...
+                else:
+                    expired = True
+                    expires = None
+
+        expired_changed = self.__expired_change(
+            self.module.get('expired'), expired)
+        expires_changed = self.__expires_change(
+            self.module.get('expires'), expires)
+
+        if expires_changed or expired_changed:
```
SlavomirMazurPantheon, line 82: Please try to change some of these condition in this file to reduce number of indentation levels.
For example if you "reverse" this condition and return False, then whole code bellow can be reduced by one level of indentation.

parseAndPopulate/resolvers/expiration.py
```diff
@@ -0,0 +1,124 @@
...
+        if self.module.get('maturity-level') == 'ratified':
+            expired = False
+            expires = None
+        if reference is not None and 'datatracker.ietf.org' in reference:
+            ref = reference.split('/')[-1]
+            rev = None
+            if ref.isdigit():
+                ref = reference.split('/')[-2]
+                rev = reference.split('/')[-1]
+            url = f'https://datatracker.ietf.org/api/v1/doc/document/?name={ref}&states__type=draft&states__slug__in=active,RFC&format=json'
```
SlavomirMazurPantheon, line 40: Move this datatracker URL to some constant

PR #613, opened 2022-09-20
added script for admin user creation

dmytro-kyrychenko:


sandbox/create_admin.py
```diff
@@ -0,0 +1,32 @@
...
+    users = RedisUsersConnection()
+
+    admin_id = users.id_by_username('admin')
+    if admin_id != b'':
+        print("Admin user already exists.")
+        exit(0)
+
+    # Getting confd-credentials from yangcatalog config
+    confd_creds = None
+    with open(os.environ['YANGCATALOG_CONFIG_PATH'], 'r') as f:
```
SlavomirMazurPantheon, line 15: Please use `create_config(config_path)` method call - you can find it in almost each script, just check how it works. Then you are able to get value from config by key.

sandbox/create_admin.py
```diff
@@ -0,0 +1,32 @@
+import os
+from redisConnections.redis_users_connection import RedisUsersConnection
+
+
+if __name__ == '__main__':
```
SlavomirMazurPantheon, line 5: Preferably, put following code at the bottom of the file:
```
if __name__ == '__main__':
    main()
```
then put the functionality into `main()` method

sandbox/create_admin.py
```diff
@@ -0,0 +1,32 @@
...
+    users = RedisUsersConnection()
+
+    admin_id = users.id_by_username('admin')
+    if admin_id != b'':
+        print("Admin user already exists.")
```
SlavomirMazurPantheon, line 10: Single quotes

PR #621, opened 2022-09-29
fix: renamed all scriptConf arguments to script_conf

Fagtoy:
renamed all scriptConf arguments to script_conf in order to generalize script arguments to avoid errors with calling scripts with an invalid argument

### DIFF
--- a/parseAndPopulate/resolve_expiration.py
+++ b/parseAndPopulate/resolve_expiration.py
@@ -27,7 +27,6 @@ __email__ = 'miroslav.kovac@pantheon.tech'
 
 import os
 import time
-import typing as t
 
 import requests
 
@@ -51,7 +50,7 @@ class ScriptConfig(BaseScriptConfig):
         super().__init__(help, None, None if __name__ == '__main__' else [])
 
 
-def main(script_conf: t.Optional[BaseScriptConfig] = None):
+def main(script_conf: BaseScriptConfig = ScriptConfig()):
     start_time = int(time.time())
 
     config = create_config()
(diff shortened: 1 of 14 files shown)


parseAndPopulate/resolve_expiration.py
```diff
@@ -55,10 +51,8 @@ def __init__(self):
         super().__init__(help, None, None if __name__ == '__main__' else [])
 
 
-def main(scriptConf=None):
+def main(script_conf: t.Optional[BaseScriptConfig] = None):
```
richardzilincikPantheon, line 54: If we never initialize a `ScriptConfig` object the `--help` argument will be ignored.

parseAndPopulate/reviseSemver.py
```diff
@@ -86,10 +86,8 @@ def load_from_json(path: str):
         return json.load(reader)
 
 
-def main(scriptConf=None):
+def main(script_conf: t.Optional[BaseScriptConfig] = None):
```
richardzilincikPantheon, line 89: same as above

utility/reviseTreeType.py
```diff
@@ -47,10 +48,8 @@ def __init__(self):
         super().__init__(help, None, [])
 
 
-def main(scriptConf=None):
+def main(script_conf: t.Optional[BaseScriptConfig] = None):
```
richardzilincikPantheon, line 51: same here

richardzilincikPantheon (changes requested):

parseAndPopulate/resolve_expiration.py
Fagtoy: thanks, fixed

richardzilincikPantheon (approved):

SlavomirMazurPantheon (approved):

PR #626, opened 2022-10-13
Sending emails to devs about failed github action

dmytro-kyrychenko:


richardzilincikPantheon (commented):
We need to agree on a maximum acceptable line length and stick to it.
Needlessly splitting function calls over multiple lines harms readability and make it harder to see who committed the original line with git blame.
I don't think we have any sort of written down style guide anywhere so I'll create a separate issue for this in YangCatalog/deployment.

SlavomirMazurPantheon:
Some changes are definitely needed, because the code review is quite complicated now. Number of lines are marked as changed, while only a change is formatting.

SlavomirMazurPantheon (approved):

PR #634, opened 2022-10-21
fix: use resolver classes in statistics.py

richardzilincikPantheon:
The organization of a module was resolved differently if an organization couldn't be determined from the module and in some other edge cases.

### DIFF
--- a/statistic/statistics.py
+++ b/statistic/statistics.py
@@ -145,7 +145,7 @@ def get_total_and_passed(directory: str) -> t.Tuple[int, int]:
             revision = '1970-01-01'
         belongs_to = BasicResolver(parsed_yang, 'belongs_to').resolve()
         namespace = NamespaceResolver(parsed_yang, LOGGER, f'{name}@{revision}', belongs_to).resolve()
-        organization = OrganizationResolver(parsed_yang, LOGGER, namespace)
+        organization = OrganizationResolver(parsed_yang, LOGGER, namespace).resolve()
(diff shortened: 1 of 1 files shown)


statistic/statistics.py
```diff
@@ -137,36 +140,22 @@ def get_total_and_passed(directory: str) -> t.Tuple[int, int]:
...
-                status = data['yang-catalog:modules']['module'][0].get('compilation-status')
-                if org == organization and rev == revision:
-                    if 'passed' == status:
-                        passed += 1
-                    num_in_catalog += 1
-            else:
-                LOGGER.error(f'Could not make request on path {path}')
+        belongs_to = BasicResolver(parsed_yang, 'belongs_to').resolve()
+        namespace = NamespaceResolver(parsed_yang, LOGGER, f'{name}@{revision}', belongs_to).resolve()
+        organization = OrganizationResolver(parsed_yang, LOGGER, namespace)
```
SlavomirMazurPantheon, line 148: calling `.resolve()` method is missing

statistic/statistics.py
```diff
@@ -137,36 +140,22 @@ def get_total_and_passed(directory: str) -> t.Tuple[int, int]:
         results = parsed_yang.search('revision')
```
SlavomirMazurPantheon, line 140: could we just call `RevisionResolver` for this?

SlavomirMazurPantheon (approved):

PR #636, opened 2022-10-25
chore: logging changes

richardzilincikPantheon:


SlavomirMazurPantheon (commented):
Maybe we should rename `LOGGER` variable for `logger` or better to say unify the name of this variable across the repository.

SlavomirMazurPantheon (approved):

PR #642, opened 2022-10-31
feat: replaced setUp methods with setUpClass where possible

### LINKED ISSUE #641: Use ```setUpClass``` and ```tearDownClass``` methods instead of ```setUp``` and ```tearDown``` methods in tests
Now there are some tests that can use ```setUpClass``` and ```tearDownClass``` methods instead of ```setUp``` and ```tearDown``` to try to speed them up, as we can, for example, avoid the creation of the config instance for every test method and create only one instance for the whole class instead. I also think there's no point in using ```___init ___``` and ```setUp``` methods in the same test class as they both execute on every test in this test class, so it would be better to keep only ```setUp``` method.

Fagtoy:
replaced ```setUp``` and ```tearDown``` methods with ```setUpClass``` and ```tearDownClass``` methods respectively, moved code from ```__init__``` methods to ```setUp``` or ```setUpClass``` methods, and refactored formatting using f-strings in all tests.

resolves #641

### DIFF
--- a/tests/test_modules.py
+++ b/tests/test_modules.py
@@ -27,7 +27,6 @@ from parseAndPopulate.models.schema_parts import SchemaParts
 from parseAndPopulate.modules import SdoModule, VendorModule
 
 
-# TODO: we should probably have unit tests for the individual Model._resolve* methods
 class TestModulesClass(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
--- a/tests/test_modules.py
+++ b/tests/test_modules.py
@@ -27,19 +27,17 @@ from parseAndPopulate.models.schema_parts import SchemaParts
 from parseAndPopulate.modules import SdoModule, VendorModule
 
 
+# TODO: we should probably have unit tests for the individual Model._resolve* methods
 class TestModulesClass(unittest.TestCase):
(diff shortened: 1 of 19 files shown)


richardzilincikPantheon (commented):
Please merge the develop branch into this to fix the gevent issue in github actions so we can actually see the results.

richardzilincikPantheon:
oh sorry, I thought https://github.com/YangCatalog/backend/pull/642 was merged

richardzilincikPantheon:
done

Fagtoy:
> done

merged

tests/test_modules.py
```diff
@@ -27,19 +27,17 @@
 from parseAndPopulate.modules import SdoModule, VendorModule
 
 
+# TODO: we should probably have unit tests for the individual Model._resolve* methods
```
richardzilincikPantheon, line 30: This isn't relevant anymore since the resolves were moved to independent classes. Please remove it to avoid confusion.

richardzilincikPantheon (changes requested):

richardzilincikPantheon (approved):

SlavomirMazurPantheon (approved):

PR #648, opened 2022-11-04
refactor: use elasticsearch aliases

richardzilincikPantheon:
Make use of elasticsearch aliases and add scripts to administrate them.
Automatic creation of missing indices is removed from 2 scripts.
An index missing is a serious error that isn't fixed just by creating
a fresh empty index.


richardzilincikPantheon:
Instructions for deploying #625:
before deploying

1. copy the following files into the running backend container:
- `elasticsearchIndexing/json/initialize_yindex_index.json`
- `sandbox/create_indices.py`
- `sandbox/reindex.py`
2. download [this script](https://gist.github.com/richardzilincikPantheon/54eaeba57c53fcbe23520f24af80da72) and copy it into `sandbox/add_es_aliases.py`
3. open a shell in the container and run
```sh
python sandbox/create_indices.py --index yindex-v2 --schema elasticsearchIndexing/json/initialize_yindex_index.json
python sandbox/reindex.py yindex yindex-v2
python sandbox/add_es_aliases.py
```

after deploying

4. `python sandbox/swap_es_aliases.py yindex yindex-v2 --delete`

SlavomirMazurPantheon:
It looks good, but I am thinking whether `sandbox/create_indices.py` script should not be moved under `elasticsearchIndexing/` directory.

richardzilincikPantheon:
> It looks good, but I am thinking whether `sandbox/create_indices.py` script should not be moved under `elasticsearchIndexing/` directory.

That's where I put it originally, but my reasoning for putting it in sandbox is that it will only be used for maintenance tasks and it's not something that should need to be used regularly. In production it will only be used when reindexing again. The other scripts in `elasticsearchIndexing` are all part of regular cronjobs.

SlavomirMazurPantheon:
True, but it is related to Elasticsearch indexing and creating the indices is pretty important.
I think that running this script should be included into setup env steps, and I am not sure whether we want to point someone to run some script in the sandbox dir.

richardzilincikPantheon:
Fair enough. I'll move it back.

richardzilincikPantheon:
Sorry, got distracted and forgot about this.

PR #652, opened 2022-11-09
Added coverage report

### LINKED ISSUE #650: Generate code coverage automatically
Update the coverage badge and generate a full coverage report summary on every Pull Request

### LINKED ISSUE #653: Update used Github Actions
Github Actions show several warnings when running tests
> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/setup-python, actions/setup-python, actions/checkout

> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

> The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see https://github.com/actions/virtual-environments/issues/6002

Outdated versions of used actions should be updated and the environment should be switched to a supported version.

Fagtoy:
I also updated GitHub Actions versions and environment and added caching of dependencies inside the ```setup-python``` step, so for now installing of dependencies is processed a little bit faster as they are cached. The ```Post coverage comment``` still gives this warning: ```Node.js 12 actions are deprecated...``` but we can't actually do anything about it for now because we should just wait for a new release of this action.

resolves #650 
resolves #653


Fagtoy:
> It seems this has some problem with loading the correct credentials for Github when running the tests on Actions. I tried re-running the job to see if it was just caused by another Action trying to push to the same repo simultaneously or something similar, but it failed again.

Yes, Slavomir is informed about this problem, he said that maybe re-generating of the token is needed, anyway this PR remains open for now, I’ll convert it draft

PR #655, opened 2022-11-10
chore: initialize number of replicas to 0 on all indices

### LINKED ISSUE #646: Disable replicas in prod ES
ES in prod is currently showing health status yellow, which means certain operations are not allowed. Should be fixed as described [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/high-availability-cluster-small-clusters.html#high-availability-cluster-design-one-node).

richardzilincikPantheon:
resolves #646

### DIFF
--- a/elasticsearchIndexing/json/initialize_autocomplete_index.json
+++ b/elasticsearchIndexing/json/initialize_autocomplete_index.json
@@ -8,7 +8,8 @@
         "read_only_allow_delete": "false"
       },
       "max_ngram_diff": 100,
-      "number_of_shards": 2
+      "number_of_shards": 2,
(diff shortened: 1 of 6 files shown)


richardzilincikPantheon:
I've changed the setting on all live indices on AWS and the production cluster now reports health status green. To do this locally, open a shell in the `yc-backend` container (or any other that has access to ES) and run
```sh
curl -X PUT 'yc-elasticsearch:9200/_all/_settings?pretty' -H 'Content-Type: application/json' -d'
{
  "index" : {
    "number_of_replicas" : 0
  }
}
'
```

SlavomirMazurPantheon (approved):

PR #656, opened 2022-11-11
chore: remove unused scripts from the sandbox directory

richardzilincikPantheon:
Reasoning:
- `check_all_tree_types.py`: Was used to reevaluate the tree type for all modules. As far as I can tell, this was because we want to reclassify some modules after they are no longer the latest revision of that module. This is now done by revise_tree_type.py in a daily cronjob.
- `delete_modules.py`: Delete all modules returned from a hard coded search query. I'm not sure for what purpose this was added or how it could be useful in the future. Nonfunctional as it wasn't updated to work with Redis.
- `file_exists_check.py`: Looks like a single use script.
- `find_conflicting_files.py`: Find modules with identical name/revision and conflicting hashes. This information was never used and it isn't clear how it could be useful.
- `find_duplicate_organizations.py`: Single use script, not clear why it was needed.
- `find_uppercase_organizations.py`: ditto
recreate_json_tree.py: Recreates all ytree json files. These are already created and recreated when necessary, I don't see when we'd want to recreate all of them.
- `redis_migrate_modules.py`: Move modules between databases. Single use script.
- `redis_migrate_vendors.py`: Move vendors between databases. Single use script.
- `reencode_redis_keys.py`: Change the format of Redis keys. Single use script.
- `remove_modules_dependents_based_on_remove_module_list.py`: Single use script, not clear why it was needed.
- `rsync_modules.py`: Single use script, name seemingly unrelated to functionality.

Please let me know if any of this is incorrect.


SlavomirMazurPantheon:
> Reasoning:
> 
> * `check_all_tree_types.py`: Was used to reevaluate the tree type for all modules. As far as I can tell, this was because we want to reclassify some modules after they are no longer the latest revision of that module. This is now done by revise_tree_type.py in a daily cronjob.
> * `delete_modules.py`: Delete all modules returned from a hard coded search query. I'm not sure for what purpose this was added or how it could be useful in the future. Nonfunctional as it wasn't updated to work with Redis.
> * `file_exists_check.py`: Looks like a single use script.
> * `find_conflicting_files.py`: Find modules with identical name/revision and conflicting hashes. This information was never used and it isn't clear how it could be useful.
> * `find_duplicate_organizations.py`: Single use script, not clear why it was needed.
> * `find_uppercase_organizations.py`: ditto
>   recreate_json_tree.py: Recreates all ytree json files. These are already created and recreated when necessary, I don't see when we'd want to recreate all of them.
> * `redis_migrate_modules.py`: Move modules between databases. Single use script.
> * `redis_migrate_vendors.py`: Move vendors between databases. Single use script.
> * `reencode_redis_keys.py`: Change the format of Redis keys. Single use script.
> * `remove_modules_dependents_based_on_remove_module_list.py`: Single use script, not clear why it was needed.
> * `rsync_modules.py`: Single use script, name seemingly unrelated to functionality.
> 
> Please let me know if any of this is incorrect.

Okay it looks reasonable and these scripts can be removed so sandbox directory will be cleaner.

SlavomirMazurPantheon (changes requested):

SlavomirMazurPantheon (approved):

PR #661, opened 2022-11-18
fix: fixed grep search and added "pagination" to not load all the results in one response

### LINKED ISSUE #662: Fix the grep search
For now, the grep search doesn't return any results if no organizations are provided. Some sort of pagination is also needed for the grep search endpoint because the response takes a lot of time

Fagtoy:
Fixed the grep search - there were no results with no organizations selected. Added some sort of "pagination" to the grep search endpoint, for now, the response will look like this: 
image
so we can load more results for the same search criteria if needed.
The amount of modules for one page is controlled from the config - ```Web-Section: grep-search-results-per-page```, default amount is set to 500. This amount actually means how many results we take from the grep search, but we can find zero or a few results for each of these modules in Elasticsearch, so basically, the final response's length is unknown, and ```grep-search-results-per-page``` only actually defines how many requests will be sent to Elasticsearch

resolves #662


SlavomirMazurPantheon (commented):
@Fagtoy do we need all these `.yang` files in `resources/test_search/all_modules` or they were pushed by accident?

SlavomirMazurPantheon:
This "problem" with duplicate files will be resolved later in this task: https://github.com/YangCatalog/backend/issues/663

Fagtoy:
I want to share possible problems with this type of pagination.
When a user tries to search for modules of some specific organization(s) we can give zero results for the first ```n``` pages and we only get some results in the ```n + 1``` page because we only take some particular amount of grep results despite of the organization, so these results that match some search criteria doesn't match the searching criteria of the organization.
The second problem that can be combined with the first one is that the results we get from the grep search could not be present in Elasticsearch (yet or at all). If we always notify indexing of Elasticsearch (which seems like that) in the ```populate.py``` then it's not a problem.
Maybe someone has some thoughts on how to avoid/fix these problems.

richardzilincikPantheon:
First thing that needs to be fixed, is possible command injection [here](https://github.com/YangCatalog/backend/blob/a37d48baf9cec178ef29e9e55d4ba6cd36e89f0f/api/views/yangSearch/grep_search.py#L78) (patched in prod now). I *think* escaping single quotes should be good enough. Try looking up shell string sanitization just in case I'm missing something.

I don't think running the grep search on each new page request is a good solution. Ideally, we'd store the results in a temporary file somewhere, and read more lines from there if we get a request for a new page. Redis has an option to set an expiration time for documents, though using Redis here might be overkill. Maybe we could just delete them once a day (and hope nobody is searching during that time).

> If after this filtering we get less than 20% (or whatever percentage we choose) of the Web-Section: grep-search-results-per-page amount of modules needed for the response, then we store the current results and perform one more grep search with the next page and repeat this until we have >= 20% of the needed amount of modules, then we search for these modules in Elasticsearch and return the results and a proper page number to load more modules

Is there a reason why we wouldn't just do this for 100%? Couldn't we just send request for each line one by one until we either have `grep-search-results-per-page` hits or lines run out?
In any case, we'll also need to store a "cursor", or how many lines we've already used up for the first/current page so that we don't show overlapping data on the next page. We also need to cache the responses for the previous pages, in case the user decides to go back a page.

Fagtoy:
@SlavomirMazurPantheon @richardzilincikPantheon regarding the comment from Richard above I decided to use the [flask-caching](https://flask-caching.readthedocs.io/en/latest/) library, so all the responses will be cached and we won't run a new grep search on every request. Grep search results for exact searches will also be stored in the cache so that the next page will re-use the cached grep result of a previous page but from a particular cursor. For now, I chose the ```FileSystemCache``` cache type because I'm not sure if other cache types will be able to store such an amount of information, and I also chose 1 hour as a default time for cache. About string sanitization:
```
search_string = (
            search_string.strip()
            .replace("'", '')
            .replace('`', r'\`')
            .replace(r'\\`', r'\`')
            .replace("'", r"\'")  # noqa: Q000
            .replace(r"\\'", r"\'")  # noqa: Q000
        )
```
as far as I understand this is enough, so no one would be able to run any script inside the search.
There's an example of how the grep search request\responses will look like:
image
image
image
image
And when we go back to one of the previous pages we see that the response time is low because the response is cached:
image
The one problem that I found is that when we use some special symbols in the search string then the response will add an additional backslash to each of the special symbols but I believe they can be removed before another request on the frontend, for example:
image

api/views/yangSearch/yangSearch.py
```diff
@@ -50,38 +53,68 @@ class YangSearch(Blueprint):
...
-    search_string: str = body.get('search')
-    inverted_search: bool = body.get('inverted_search', False)
-    case_sensitive: bool = body.get('case_sensitive', False)
+    if not (query_params := request.args):
+        abort(400, description='No query parameters were provided')
+    search_string: str = query_params.get('search')
     if not search_string:
         abort(400, description='Search cannot be empty')
+    organizations: list[str] = query_params.getlist('organizations')
+    inverted_search: bool = query_params.get('inverted_search', type=lambda v: v.lower() == 'true')
```
richardzilincikPantheon, line 74: This could be a `None` which would make the type hint not true. I guess this probably won't cause errors in the current code since it's still falsy, but it would be better to provide `False` as the default.

richardzilincikPantheon (changes requested):

api/views/yangSearch/grep_search.py
```diff
@@ -101,43 +100,38 @@ def _get_matching_module_names(
...
-        except subprocess.CalledProcessError as e:
-            if not e.output and inverted_search:
-                self.logger.info('All the modules satisfy the inverted search')
-                return self._get_modules_from_cursor(self._get_all_modules_with_filename_extension())
-            elif not e.output:
-                self.logger.info(f'Did not find any modules satisfying such a search: {search_string}')
-                return
+            command_output, error = search_command.communicate()
+        except Exception as e:
+            self.logger.exception(f'Such a search: {search_string}, caused an error: {str(e)}')
```
richardzilincikPantheon, line 117: Doesn't logging and then raising again mean there will be two different errors logged for this? We should avoid that if possible. The exception type should also be more specific. I think `OSError` is the only one that could occur here.

api/views/yangSearch/grep_search.py
```diff
@@ -187,15 +212,27 @@ def _get_modules_from_cursor(
         except IndexError:
             return None
 
-    def _get_all_modules_with_filename_extension(self) -> list[str]:
-        all_modules = cache.get(self.listdir_results_cache_key)
+    def _get_all_modules_with_filename_extension(self, return_set: bool = False) -> t.Union[set[str], list[str]]:
```
richardzilincikPantheon, line 215: I don't think we should complicate the return type like this. It confuses the type checker and the performance impact for the conversion is negligible (less than a ms for 10000 64 char strings on my machine).

api/views/yangSearch/grep_search.py
```diff
@@ -66,38 +90,170 @@ def _get_matching_module_names(
             :param inverted_search  (bool) indicates if the result must contain all modules satisfying the search or
             all the modules not satisfying the search
             :param case_sensitive   (bool) indicates if the search must be case-sensitive or not
+            :param organizations   (list[str]) names of organizations that modules should be of
         """
-        module_names_with_format = set()
-        pcregrep_shell_command = f'pcregrep -{"" if case_sensitive else "i"}lrMe'
+        search_string = (
```
richardzilincikPantheon, line 95: We can remove this now that the string isn't being parsed by a shell.

richardzilincikPantheon (changes requested):

api/views/yangSearch/grep_search.py
Fagtoy: The only problem I see here is that the error text would be duplicated if ```exc_info``` is set to ```True```: 
image
so I set it to ```False```, did you mean this?

richardzilincikPantheon: I mean that in yang.log we'll see the message explicitly logged here with `logger.exception()`, and flask will log it's own message when it catches the `ValueError`. We could just pass the `f'Such a search: {search_string}, caused an error: {error}'` string into the `ValueError` and let flask catch it. Using `logger.exception()` mostly makes sense when we don't want to propagate the exception.

Fagtoy: Got it, thanks, fixed

richardzilincikPantheon (approved):

SlavomirMazurPantheon (approved):

PR #668, opened 2022-11-22
feat: remove uuid directories from /var/yang/tmp in remove_unused.py script

### LINKED ISSUE #667: Remove uuid directories in temp directory
There are directories with UUID-like names that were created but then not deleted in the ```/var/yang/tmp``` dir on prod. So these directories should be deleted in the ```removeUnused.py``` script

Fagtoy:
Renamed the ```removeUnused.py``` script to ```remove_unused.py``` and added removing of UUID-like directories from the ```/var/yang/temp``` directory in this script. I also added removing the temp JSON dir in the ```populate.py``` script.

resolves #667

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -123,10 +123,10 @@
       "justMyCode": false
     },
     {
-      "name": "removeUnused",
+      "name": "remove_unused",
       "type": "python",
       "request": "launch",
-      "program": "${workspaceFolder}/utility/removeUnused.py",
+      "program": "${workspaceFolder}/utility/remove_unused.py",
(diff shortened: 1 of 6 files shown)


SlavomirMazurPantheon (approved):

richardzilincikPantheon:
Unless github isn't showing me something correctly, I think you forgot to rename the file itself.

Fagtoy:
> Unless github isn't showing me something correctly, I think you forgot to rename the file itself.

Thanks, didn't notice. I created a separate PR for this: https://github.com/YangCatalog/backend/pull/670

PR #684, opened 2022-12-12
Divide ycSearch

dmytro-kyrychenko:


richardzilincikPantheon (commented):
`create_tree()` and `create_reference()` don't seem to make use of redis despite being in redisSearch.py. Is there a reason for why they're in this file, or could they perhaps be moved to yangSearch.py? The goal is to be able to guess in which file any given piece of functionality is implemented in without having to do a full text search of the codebase.

dmytro-kyrychenko:
@richardzilincikPantheon yeah, I checked they do not use redis. Also there could be a problem with moving them to yangSearch, because yangSearch has "/api/yang-search/v2" url prefix and this redisSearch has "/api" prefix. So yeah, the question is do we want to save the same prefix for these two endpoints and if so, maybe I'll need to move them to completely separate file

richardzilincikPantheon:
Oh I see. Yeah I think we shouldn't really change the endpoint URLs. In the [docs](https://yangcatalog.org/doc/#services), these two endpoints are grouped together with some of the endpoints from `compareModules` under the "Services" section. The rest of the endpoints in `compareModules` are grouped together with the endpoints currently in `redisSearch` under the "Search" and "RPC Search" sections.

Splitting the files in the same ways would make the most sense to me, that way the whole services blueprint could have a prefix of `api/services`. Search and RPC Search endpoints can go in a single file as I don't see any kind of logical separation there. Could these two sections be merged in the documentation as well or do we want to keep them separate @SlavomirMazurPantheon?

I've also noticed that the endpoints in `yangSearch` appear to all be meant for internal use by our own frontend rather than as a public API. The name of this module should be changed to reflect that.

api/views/compareModules/compareModules.py
```diff
@@ -0,0 +1,382 @@
...
+    schema2 = '{}/{}@{}.yang'.format(ac.d_save_file_dir, name2, revision2)
+    file_name1 = 'schema1-file-diff.txt'
+    yang_file_1_content = ''
+    try:
+        with open(schema1, 'r', encoding='utf-8', errors='strict') as f:
+            yang_file_1_content = f.read()
+    except FileNotFoundError:
+        app.logger.warn('File {}@{}.yang was not found.'.format(name1, revision1))
+
+    with open('{}/{}'.format(ac.w_save_diff_dir, file_name1), 'w+') as f:
```
SlavomirMazurPantheon, line 120: Use os.path.join here instead of formatting

api/views/compareModules/compareModules.py
```diff
@@ -0,0 +1,382 @@
...
+        f.write('<pre>{}</pre>'.format(yang_file_1_content))
+
+    file_name2 = 'schema2-file-diff.txt'
+    yang_file_2_content = ''
+    try:
+        with open(schema2, 'r', encoding='utf-8', errors='strict') as f:
+            yang_file_2_content = f.read()
+    except FileNotFoundError:
+        app.logger.warn('File {}@{}.yang was not found.'.format(name2, revision2))
+    with open('{}/{}'.format(ac.w_save_diff_dir, file_name2), 'w+') as f:
```
SlavomirMazurPantheon, line 130: os.path.join

SlavomirMazurPantheon (commented):
Separation looks good, but please replace `.format()` method with f-strings everywhere and also in appropriate cases use `os.path.join` instead of creating path using `format()`

SlavomirMazurPantheon (approved):

PR #687, opened 2022-12-15
Clear all modules directory in tests

### LINKED ISSUE #663: Clear modules after running the tests
Before running tests, we run the [prepare environment](https://github.com/Fagtoy/backend/blob/develop/prepare_environment.sh) script which populates the ```tests/resources/all_modules``` directory with a lot of files (for now, after running this script there are 1255 files), and these files are not cleared up after tests. We should understand which tests really need all of these populated files, and maybe this directory can be populated with less amount of modules for a separate test class. All the created modules during the test run should be deleted after.

Fagtoy:
Refactored the ```prepare_environment.sh``` script, so it no longer runs the ```generate_schema_urls.py``` script, and the ```tests/resources/all_modules``` directory isn't populated with new modules, from now on, only test classes that need populated modules run the  ```generate_schema_urls.py``` script and clear the populated modules after. I also fixed some existing broken tests related to the ```fetch_modules``` function
**NOTE**: tests must pass after merging this PR as the workflow is changed

resolves #663


richardzilincikPantheon (changes requested):

richardzilincikPantheon (approved):

sandbox/check_schema.py
```diff
@@ -130,7 +130,7 @@ def __print_patch_response(key: str, response):
 
     # GET all the existing modules of Yangcatalog
     LOGGER.info('Fetching all of the modules from API')
-    all_existing_modules = fetch_modules(LOGGER)
+    all_existing_modules = fetch_modules(LOGGER, config=config)
```
SlavomirMazurPantheon, line 133: Why config cannot be loaded directly in `fetch_modules` method?
Like why we need to pass it as an argument when "default" argument value can be used?

Fagtoy: Just to not create a new config instance every time, there's no deeper meaning behind this. The only other case is when some variable in the config can be changed, like in tests, so we can control this

SlavomirMazurPantheon (approved):

PR #703, opened 2023-01-17
Remove script configs

dmytro-kyrychenko:


SlavomirMazurPantheon (commented):
We definitely do not want separate config file for every script - it is even worse and less readable than it was before.
Instead of this create one python file which will group these configs.
This file will contain one large dict - key in this will be name of the file, value will be config
```python
config = {
    "process_changed_mods": {
        "help": 'Process changed modules in a git repo',
        "args": [
            {
                'flag': '--config-path',
                'help': 'Set path to config file',
                'type': str,
                'default': os.environ['YANGCATALOG_CONFIG_PATH'],
            },
        ]
    },
    "...": ....
}
```

elasticsearchIndexing/process_changed_mods.py
```diff
@@ -22,30 +22,20 @@
...
 
 import requests
 
 from elasticsearchIndexing.build_yindex import build_indices
 from elasticsearchIndexing.es_manager import ESManager
 from utility import log
 from utility.create_config import create_config
-from utility.scriptConfig import Arg, BaseScriptConfig
+from utility.script_config_dict import script_config_dict
+from utility.scriptConfig import BaseScriptConfig
```
SlavomirMazurPantheon, line 33: As we do not use `ScriptConfig` class in individual scripts, then `BaseScriptConfig` can be renamed to `ScriptConfig`

elasticsearchIndexing/process_changed_mods.py
```diff
@@ -22,30 +22,20 @@
...
-                'flag': '--config-path',
-                'help': 'Set path to config file',
-                'type': str,
-                'default': os.environ['YANGCATALOG_CONFIG_PATH'],
-            },
-        ]
-        super().__init__(help, args, None if __name__ == '__main__' else [])
+help = script_config_dict['process_changed_mods']['help']
+args = script_config_dict['process_changed_mods']['args']
+DEFAULT_SCRIPT_CONFIG = BaseScriptConfig(help, args, None if __name__ == '__main__' else [])
```
SlavomirMazurPantheon, line 38: This can be refactored into something like:
```python
BASENAME = os.path.basename(__file__)
FILENAME = BASENAME.split('.py')[0]

DEFAULT_SCRIPT_CONFIG = BaseScriptConfig(
        help=script_config_dict[FILENAME]['help'], 
        args=script_config_dict[FILENAME]['args'], 
        None if __name__ == '__main__' else [])
```

ietfYangDraftPull/draftPull.py
```diff
@@ -41,35 +40,22 @@
 from ietfYangDraftPull import draftPullUtility as dpu
 from utility import message_factory
 from utility.create_config import create_config
-from utility.scriptConfig import Arg, BaseScriptConfig
+from utility.script_config_dict import script_config_dict
+from utility.scriptConfig import ScriptConfig
 from utility.util import job_log
 
+BASENAME = os.path.basename(__file__)
```
SlavomirMazurPantheon, line 47: In some files there is already line of code `current_file_basename = os.path.basename(__file__)` which is same as `BASENAME = os.path.basename(__file__)`
Please use only one of these variables - I would prefer `BASENAME`

utility/script_config_dict.py
```diff
@@ -0,0 +1,430 @@
...
+            },
+            {
+                'flag': '--dir',
+                'help': 'Set directory where to look for hello message xml files',
+                'type': str,
+                'default': '/var/yang/nonietf/yangmodels/yang/standard/ietf/RFC',
+            },
+            {'flag': '--output', 'help': 'Output json file', 'type': str, 'default': 'integrity.json'},
+        ],
+        'yang_models': config.get('Directory-Section', 'yang-models-dir'),
```
richardzilincikPantheon, line 121: I think this isn't meant to be here.

utility/script_config_dict.py
```diff
@@ -0,0 +1,430 @@
...
+
+from utility.create_config import create_config
+from utility.staticVariables import backup_date_format
+
+config = create_config()
+credentials = config.get('Secrets-Section', 'confd-credentials').strip('"').split()
+save_file_dir = config.get('Directory-Section', 'save-file-dir')
+result_dir = config.get('Web-Section', 'result-html-dir')
+
+script_config_dict = {
```
richardzilincikPantheon, line 31: Try and give this a type hint, so the type checker can check that all the args are in the correct format. You might need a new `TypedDict` class to make it work. Without checks, editing a blob this big could lead to a lot of mistakes.

richardzilincikPantheon (changes requested):

utility/scriptConfig.py
```diff
@@ -103,3 +118,6 @@ def get_args_list(self) -> t.Dict:
 
     def get_help(self) -> Help:
         return self.help
+
+    def copy(self):
+        return ScriptConfig(**self.copy_info)
```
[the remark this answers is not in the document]
dmytro-kyrychenko: No, it wouldn't work, because when ScriptConfig is initialized, it changes the initialization data that is passed to it. The example is args.pop('flag') in __add_args. So, if we use the data, that was already processed by first ScriptConfig initialization, then we will have errors on second ScriptConfig initialization (no 'flag' in args already)

utility/script_config_dict.py
dmytro-kyrychenko: That info was inside previous ScriptConfig for this file, I checked it is not used in the integrity.py, and it works without it, so yeah, I guess I can remove it

utility/scriptConfig.py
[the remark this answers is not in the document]
richardzilincikPantheon: Ah ok, I see.

richardzilincikPantheon (changes requested):
Just one more thing, make a base class with just the required keys, and subclass it with `total=False` for the optional keys, similar to how it's done with `Arg` [here](https://github.com/YangCatalog/backend/blob/06e7a7ddbc34f417f7fc4798b38502265ceeedcf/utility/scriptConfig.py#L29). That way we don't need the `None`s. Python 3.11 has more convenient syntax for this that let's you mark keys as `NotRequired[...]`, so we can get rid of the subclassing in the future after we upgrade.

richardzilincikPantheon (approved):

SlavomirMazurPantheon (approved):

PR #705, opened 2023-01-19
Distinguish rfc and not rfc modules during search

Fagtoy:
Added the possibility to search only for ratified modules during the modules search. To apply changes in PROD:

```python sandbox/update_es_index_mapping.py --index yindex-alias --new-mapping-path sandbox/update_yindex_index.json```

```python populate_maturity_level_in_es_yindex.py``` (here is code of this script: https://gist.github.com/Fagtoy/353c009b90369e017b3577eab103198f)

### CODE REVIEW

File: sandbox/populate_maturity_level_in_es_yindex.py [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+        es_response = es_manager.generic_search(index=ESIndices.YINDEX, query=query, response_size=None)['hits']['hits']
+        for result in es_response:
+            module_data = result['_source']
+            module_data['maturity-level'] = maturity_level
+            modules_for_bulk_indexing.append(module_data)
+    logger.info(f'{len(modules_for_bulk_indexing)} modules are ready to be updated in Elasticsearch')
+    i = 0
+    while i <= len(modules_for_bulk_indexing):
+        chunk = modules_for_bulk_indexing[i : i + ES_CHUNK_SIZE]
+        es_manager.bulk_modules(ESIndices.YINDEX, chunk)
```
richardzilincikPantheon, line 51: Does this really just update the documents? I think this would add a second version of the document, so the one without the maturity-level will stay in the index as well. I don't think you can modify the source of document after it's been indexed. Either you need to reindex, or use the [update API](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html). I think the update API let's you set properties without modifying the source.

Fagtoy: thanks, changed to update API

sandbox/update_yindex_index.json
```diff
@@ -0,0 +1,50 @@
...
+        }
+      },
+      "sdo": {
+        "type": "boolean"
+      },
+      "revision": {
+        "type": "date"
+      },
+      "maturity-level": {
+        "type": "keyword"
```
SlavomirMazurPantheon, line 47: Do we need to store info about maturity-level?
I mean like whether is it not enough to store just some `rfc` flag - similar as for `sdo`
```
      "rfc": {
        "type": "boolean"
      }
```

Fagtoy: I’m not sure, for this purpose it would be enough but maybe we would have to filter/search for some specific maturity-level in future, as with sdo, we still store the organization and just have another boolean field

api/views/yangSearch/response_row.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, elastic_hit: dict) -> None:
         self.origin = self._set_origin(elastic_hit['organization'])
         self.organization = elastic_hit['organization']
         self.description = elastic_hit['description']
-        self.maturity = elastic_hit.get('maturity-level', '')
+        self.maturity = 'ratified' if elastic_hit.get('rfc') else ''
```
SlavomirMazurPantheon, line 37: Maybe it would be enough to use only: 
`self.rfc = elastic_hit.get('rfc')`
and work only with this True/False value
Or does we need this value explicitly somewhere?

Fagtoy: This is the output row, so we show it in the response

sandbox/populate_maturity_level_in_es_yindex.py
Fagtoy: https://gist.github.com/Fagtoy/353c009b90369e017b3577eab103198f
this script was removed from the repo

richardzilincikPantheon: OK, you need to check the revision when setting rfc to true though. An older revision of a module might have been published in a draft and a later one in an rfc. The revision from the draft shouldn't be marked as ratified.

Fagtoy: > OK, you need to check the revision when setting rfc to true though. An older revision of a module might have been published in a draft and a later one in an rfc. The revision from the draft shouldn't be marked as ratified.

how can I check it correctly?

Fagtoy: Updated the [gist](https://gist.github.com/Fagtoy/353c009b90369e017b3577eab103198f)

richardzilincikPantheon (approved):

SlavomirMazurPantheon (approved):

PR #711, opened 2023-01-25
fix: hash password before storing to Redis

SlavomirMazurPantheon:
Signed-off-by: SlavomirMazurPantheon <slavomir.mazur@pantheon.tech>

### DIFF
--- a/sandbox/create_admin.py
+++ b/sandbox/create_admin.py
@@ -1,20 +1,26 @@
+import hashlib
+
 from redisConnections.redis_users_connection import RedisUsersConnection
 from utility.create_config import create_config
+
+
+def hash_pw(password: str) -> str:
+    encoded_password = password.encode(encoding='utf-8', errors='strict')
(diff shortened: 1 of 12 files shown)


sandbox/create_admin.py
```diff
@@ -1,20 +1,26 @@
+import hashlib
+
 from redisConnections.redis_users_connection import RedisUsersConnection
 from utility.create_config import create_config
 
 
+def hash_pw(password: str) -> str:
```
Fagtoy, line 7: we have this function https://github.com/YangCatalog/backend/blob/develop/api/authentication/auth.py#L17 which hashes the password before it gets checked, so I believe we shouldn't have two functions for hashing the password because if one changes then another one should be changed to, maybe it would be better to place this function somewhere else and use it in this script and in auth.py as well?

Fagtoy (changes requested):

Fagtoy (approved):

PR #718, opened 2023-02-01
Added backend endpoint for visual distinction between ratified …

dmytro-kyrychenko:
…and non-ratified revisions


SlavomirMazurPantheon:
@dmytro-kyrychenko try to investigate whether it would be possible to extend functionality in existing API endpoint:
`https://yangcatalog.org/api/yang-search/v2/module-details/ietf-routing`
One of its properties is `revisions` which is list of strings (string = revision in this case).
Maybe it should return list of revisions, with flag whether it is or is not a ratified module
```json
{
    "current-module": "ietf-routing@2018-03-13.yang",
    "metadata": {},
    "revisions": [
    	{"2018-03-13", true}
    	{"2018-01-25", false}
    ]
}
```
I think. that list of revisions if currently fetched from ElasticSearch and since previous release we are also indexing info about maturity-level into Elastic so it might worth checking.

richardzilincikPantheon:
> and since previous release we are also indexing info about maturity-level into Elastic

We do, but not in the autocomplete index. I think it wouldn't be a problem to search the yindex instead, in fact that would probably work everywhere we currently use the autocomplete index.
Getting the info directly from ES would also mean we don't have to make a request to redis for every revision of the module.

api/views/yangSearch/yangSearch.py
```diff
@@ -446,6 +448,46 @@ def module_details(module: str, revision: t.Optional[str] = None, warnings: bool
     return response
 
 
+@bp.route('/get-revisions-maturity-level/<module>', methods=['GET'])
```
SlavomirMazurPantheon, line 451: is this endpoint still needed?

api/views/yangSearch/yangSearch.py
```diff
@@ -569,6 +611,28 @@ def get_modules_revision_organization(module_name: str, revision: t.Optional[str
         abort(404, f'Failed to get revisions and organization for {name_rev}')
 
 
+def get_modules_revision_maturity_levels(module_name: str):
```
SlavomirMazurPantheon, line 614: define return value of method

SlavomirMazurPantheon (approved):

elasticsearchIndexing/es_manager.py
```diff
@@ -282,6 +282,19 @@ def get_sorted_module_revisions(self, index: ESIndices, name: str):
 
         return es_result['hits']['hits']
 
+    def get_maturity_levels_for_module(self, module_name: str):
```
richardzilincikPantheon, line 285: Doesn't this just duplicate the functionality of `get_sorted_module_revisions()`?

api/views/yangSearch/yangSearch.py
```diff
@@ -569,6 +571,28 @@ def get_modules_revision_organization(module_name: str, revision: t.Optional[str
         abort(404, f'Failed to get revisions and organization for {name_rev}')
 
 
+def get_modules_revision_maturity_levels(module_name: str) -> list[dict[str, str]]:
```
richardzilincikPantheon, line 574: This is very similar to `get_modules_revision_organization()`. Instead of adding a new function, `get_modules_revision_organization()` could return this `revision: rfc` dict instead of the list of revision.

richardzilincikPantheon (changes requested):

elasticsearchIndexing/es_manager.py
dmytro-kyrychenko: not really, because get_sorted_module_revisions() gets info only about one module, so if I use it we are making many requests to elastic again like before

api/views/yangSearch/yangSearch.py
dmytro-kyrychenko: yeah, you are right, I'll change that

elasticsearchIndexing/es_manager.py
dmytro-kyrychenko: I looked one more time, I think if I change ESIndex for old methods, I can find rfc without new methods

richardzilincikPantheon (approved):

api/views/yangSearch/yangSearch.py
```diff
@@ -550,16 +550,22 @@ def get_modules_revision_organization(module_name: str, revision: t.Optional[str
...
 
         organization = hits[0]['_source']['organization']
         revisions = []
         for hit in hits:
             hit = hit['_source']
-            revisions.append(hit['revision'])
+            revision = hit['revision']
+            revision_mat_level = {
+                'revision': revision,
+                'maturity_level': hit['rfc'] if 'rfc' in hit else False,
```
SlavomirMazurPantheon, line 565: Please either name this property `maturity_level` and return its value (like ratified, adopted ...) OR name this property `is_rfc` and then return True/False.

SlavomirMazurPantheon (approved):

PR #723, opened 2023-02-06
Extract code snippets

Fagtoy:
added the endpoint to view all code snippets of a draft

### DIFF
--- a/api/views/yangSearch/yangSearch.py
+++ b/api/views/yangSearch/yangSearch.py
@@ -443,6 +443,23 @@ def module_details(module: str, revision: t.Optional[str] = None, warnings: bool
     return response
 
 
+@bp.route('/draft-code-snippets/<draft_name>', methods=['GET'])
+def get_draft_code_snippets(draft_name: str) -> t.Optional[list[str, ...]]:
+    config = create_config()
+    code_snippets_directory = config.get('Web-Section', 'code-snippets-directory')
+    draft_code_snippets_directory = os.path.join(code_snippets_directory, os.path.splitext(draft_name)[0])
+    if not os.path.exists(draft_code_snippets_directory):
+        return jsonify([])
+    domain_prefix = config.get('Web-Section', 'domain-prefix')
+    public_directory = config.get('Web-Section', 'public-directory')
+    draft_code_snippets_directory_relpath = os.path.relpath(draft_code_snippets_directory, public_directory)
+    draft_code_snippets_url = f'{domain_prefix}/{draft_code_snippets_directory_relpath}'
+    response = []
+    for filename in os.listdir(draft_code_snippets_directory):
+        response.append(f'{draft_code_snippets_url}/{filename}')
+    return jsonify(response)
+
+
 @bp.route('/yang-catalog-help', methods=['GET'])
 def get_yang_catalog_help():
     """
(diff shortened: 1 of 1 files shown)


api/views/yangSearch/yangSearch.py
```diff
@@ -445,6 +445,28 @@ def module_details(module: str, revision: t.Optional[str] = None, warnings: bool
     return response
 
 
+@bp.route('/draft-code-snippets/<draft_name>', methods=['GET'])
+def get_draft_code_snippets(draft_name: str) -> t.Optional[list[str, ...]]:
```
richardzilincikPantheon, line 449: `list[str, ...]` -> `list[str]`
Also please add a documentation comment.

richardzilincikPantheon (changes requested):

api/views/yangSearch/yangSearch.py
richardzilincikPantheon: I realized `jsonify()` actually returns a string, so it should probably just be `t.Optional[str]`.

api/views/yangSearch/yangSearch.py
```diff
@@ -445,6 +445,28 @@ def module_details(module: str, revision: t.Optional[str] = None, warnings: bool
...
+    )
+    draft_code_snippets_directory = os.path.join(code_snippets_directory, os.path.splitext(draft_name)[0])
+    if not os.path.exists(draft_code_snippets_directory):
+        return jsonify([])
+    domain_prefix = config.get('Web-Section', 'domain-prefix')
+    public_directory = config.get('Web-Section', 'public-directory')
+    draft_code_snippets_directory_relpath = os.path.relpath(draft_code_snippets_directory, public_directory)
+    draft_code_snippets_url = f'{domain_prefix}/{draft_code_snippets_directory_relpath}'
+    response = []
+    for filename in os.listdir(draft_code_snippets_directory):
```
SlavomirMazurPantheon, line 465: This can be possibly refactored to
```
    response = [
        f'{draft_code_snippets_url}/{filename}'
        for filename in os.listdir(draft_code_snippets_directory)
    ]
```

api/views/yangSearch/yangSearch.py
```diff
@@ -445,6 +445,28 @@ def module_details(module: str, revision: t.Optional[str] = None, warnings: bool
     return response
 
 
+@bp.route('/draft-code-snippets/<draft_name>', methods=['GET'])
+def get_draft_code_snippets(draft_name: str) -> t.Optional[list[str, ...]]:
+    config = create_config()
```
SlavomirMazurPantheon, line 450: Cannot we use `app_config` global variable?

api/views/yangSearch/yangSearch.py
Fagtoy: I found out that ```jsonify()``` actually returns a ```Reponse``` object, fixed

richardzilincikPantheon: Ah, ok. Do add a docstring though, ideally all the endpoints where the response format isn't immediately obvious from the name should have one.

Fagtoy: Sorry, forgot about documentation, added now

richardzilincikPantheon (approved):

SlavomirMazurPantheon (approved):

PR #725, opened 2023-02-09
feat: added automatic PR creation for new RFC

Fagtoy:
Added automatic PullRequest creation to the: https://github.com/YangModels/yang/ repo if there are new/different files found in our forked repository. The result of the PR creation is then included in the message about new files
Screenshot from 2023-02-09 15-00-38


automatic_push/rfc_push.py
```diff
@@ -0,0 +1,136 @@
...
+            continue
+        shutil.copy2(extracted_file_path, filename)
+        if os.path.islink(filename_without_revision):
+            os.unlink(filename_without_revision)
+        os.symlink(filename, filename_without_revision)
+    os.chdir(cwd)
+    shutil.rmtree(extract_to)
+
+
+def _create_pull_request(
```
SlavomirMazurPantheon, line 103: I would also think about moving part of this method into repoutil. At least the code responsible for creating PR.
Try to compare it with code in `draftPull.py` local and create method which will get necessary info as parameter

SlavomirMazurPantheon:
Also please add README.md file in the new `automatic_push` directory.

richardzilincikPantheon (changes requested):

richardzilincikPantheon (approved):

automatic_push/rfc_push.py
```diff
@@ -0,0 +1,124 @@
...
+        'https://api.github.com/repos/YangModels/yang/pulls',
+        headers={'Authorization': f'token {token}', 'Content-Type': 'application/vnd.github+json'},
+        data=json.dumps({'state': 'open', 'head': f'yang-catalog:{forked_repo.repo.active_branch}'}),
+    )
+    if prs := open_pr_response.json():
+        return PullRequestCreationResult(
+            False,
+            f'There is already an open PullRequest (files were updated in this PullRequest): {prs[0]["html_url"]}',
+        )
+    response = create_pull_request(
```
SlavomirMazurPantheon, line 114: what about passing these parameters as one dictionary object
I can see they consist mostly of strings and other dicts.
You can create some dataclass with name `PullRequestDetail` for example and pass it as 1 parameter.

richardzilincikPantheon: What would the advantage of this be here?

SlavomirMazurPantheon: Instead of passing 4 strings as parameters + one dict which contains headers, everything will be in one object.

Fagtoy: I created a dataclass for this purpose for the future (for example, ycJobs.py can use this function) but for now, I removed the manual creation of a PR because a Pr would anyway be created after the successful run of the GitHub Actions, so files will only be pushed to the repo

richardzilincikPantheon: > Instead of passing 4 strings as parameters + one dict which contains headers, everything will be in one object.

That's true, but the call to the constructor for the dataclass will still need to be called with all 5 of the parameters. I don't really see how that improves the situation. It would make sense if there were multiple places where we'd need the same set of arguments or if they all needed to be passed through several levels of functions, but that's not the case here.

SlavomirMazurPantheon: You are absolutely right, I didn't express myself properly. Instead of `dataclass`, I was thinking of `TypedDict`.

Fagtoy: I think it would be more beneficial to use a ```TypedDict``` only for the ```request_body``` param because it's not immediately understandable (moreover, when there are functions for creating, approving, merging, ...)
So, there will be ```TypedDicts``` for the request body of each of these functions, what do you think?

ietfYangDraftPull/draftPull.py
```diff
@@ -116,11 +117,22 @@ def main(script_conf: ScriptConfig = DEFAULT_SCRIPT_CONFIG.copy()) -> list[dict[
                 os.chmod(rfc_exceptions, 0o664)
                 remove_from_new = []
             new_files = [file_name for file_name in new_files if file_name not in remove_from_new]
-
+            automatic_pull_request_creation_result = create_new_rfcs_pull_request(
```
SlavomirMazurPantheon, line 120: As we discussed earlier we will just create and push the commit to the repo. API endpoint is responsible for creating Pull Request, so we should not use `pull_request` term here at all.

SlavomirMazurPantheon (approved):

PR #726, opened 2023-02-15
chore: remove logic for maintaining GitHub schema URLs

richardzilincikPantheon:
YANG files are instead assumed to be hosted at YC.O


richardzilincikPantheon:
This script replaces all the existing URLs in the redis database. https://gist.github.com/richardzilincikPantheon/0f9632a8bd848fb8a997dcc6dabb8dc0

PR #730, opened 2023-02-17
Etsi populate

dmytro-kyrychenko:


richardzilincikPantheon:
@dmytro-kyrychenko is this PR ready for review? If so please mark it as such.

richardzilincikPantheon:
I realized that after my changes in #743, populating modules from the additional repos could simply be done with adding the URLs to the repos [here](https://github.com/YangCatalog/module-compilation/blob/develop/cronjobs/run_download_github.sh#L16) and adding the corresponding directories [here](https://github.com/YangCatalog/backend/pull/743/files#diff-b487770d3608748bfc025a94911785bbd26984cf1f69c55a480904df2b42faa3R126).

dmytro-kyrychenko:
@richardzilincikPantheon @SlavomirMazurPantheon so, should I remove this changes and add those urls to the repo?

richardzilincikPantheon:
I think so, yeah.

SlavomirMazurPantheon:
Someone should test whether it will be working correctly when we add `'standard/etsi'` [here](https://github.com/YangCatalog/backend/pull/743/files#diff-b487770d3608748bfc025a94911785bbd26984cf1f69c55a480904df2b42faa3R126)
If yes, then we do not need to have separate script, which will triggers this populate.

richardzilincikPantheon:
Superseded by #743

PR #732, opened 2023-02-21
chore: remove feb 29 replacement

richardzilincikPantheon:
In some cases, replacing feb 29 with feb 28 was just incorrect. In general, writing feb 29 instead of feb 28 isn't more common than any other nonsensial mistake, so we don't have a reason to special case it.

### DIFF
--- a/api/views/yangSearch/elkSearch.py
+++ b/api/views/yangSearch/elkSearch.py
@@ -200,7 +200,7 @@ class ElkSearch:
             if module_key in reject:
                 continue
 
-            module_latest_revision = self._latest_revisions.get(row.module_name, '').replace('02-29', '02-28')
+            module_latest_revision = self._latest_revisions.get(row.module_name, '')
             if self._search_params.latest_revision and row.revision != module_latest_revision:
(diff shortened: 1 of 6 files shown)


SlavomirMazurPantheon:
@richardzilincikPantheon I think the main reason we were replacing this date was that it raised an error when we tried to pass module with this date to the Elasticsearch or ConfD (I am not sure which one).
Please, if you not already, test this scenario properly.

Fagtoy (approved):

richardzilincikPantheon:
> @richardzilincikPantheon I think the main reason we were replacing this date was that it raised an error when we tried to pass module with this date to the Elasticsearch or ConfD (I am not sure which one). Please, if you not already, test this scenario properly.

I didn't think of that, will do. If it turns out to be confd, then I'll just move the replacement to the classes working with confd and take out the rest of it.

richardzilincikPantheon:
From my testing, it seems confd only checks that dates have the correct number of digits, and elasticsearch has full support for leap years, so it shouldn't be an issue.

PR #734, opened 2023-02-21
feat: moved draftPull and ianaPull scripts in the automatic_push directory

### LINKED ISSUE #731: Move ```ianaPull.py``` and ```draftPull.py``` scripts to the ```automatic_push``` directory
There's a new ```automatic_push``` directory which should include all the scripts related to pushing the new data to GitHub, so the ```ianaPull.py``` and ```draftPull.py``` scripts should be moved into this directory

Fagtoy:
Moved ```draftPull.py``` and ```ianaPull.py``` in the ```automatic_push```, refactored these scripts as classes with methods so they became more readable, moved a few functions from ```draftPullUtility.py``` to ```repoutil.py``` and ```automatic_push/utility.py```, created ```push_untracked_files``` function in ```automatic_push/utility.py``` because this function is used in ```draftPull.py``` and ```ianaPull.py```

resolves #731


automatic_push/utility.py
```diff
@@ -0,0 +1,165 @@
...
+        # git push fork main
+        push_info = fork.push('main')[0]
+        logger.info(f'Push info: {push_info.summary}')
+        if 'non-fast-forward' in push_info.summary:
+            logger.warning('yang-catalog/yang repo might not be up-to-date')
+    except GitCommandError:
+        logger.exception('yang-catalog/yang repo might not be up-to-date')
+
+
+def populate_draft_modules_content(experimental_path: str, config: ConfigParser, logger: logging.Logger):
```
richardzilincikPantheon, line 77: Same as above, I'd avoid "populate". Maybe download or copy instead?

utility/repoutil.py
```diff
@@ -42,26 +44,36 @@ class RepoUtil:
...
+        clone_options: t.Optional[CloneOptions] = None,
         logger: t.Optional[logging.Logger] = None,
     ):
         """
         Arguments:
             :param repourl          (str) URL of the repository
             :param clone            (bool) Should always be set to True. To load a repository
                                            which has already been cloned, see  the load() function.
-            :param clone_options    (dict) May contain the keys local_dir, config_username and config_email
+            :param clone_options    (Optional[dict]) May contain the information for the repository cloning
```
richardzilincikPantheon, line 69: `Optional[CloneOptions]`

richardzilincikPantheon (changes requested):

README.md
```diff
@@ -88,14 +88,15 @@ and their expiration dates and updates its metadata accordingly.
 anymore.
 * [user_reminder](https://github.com/YangCatalog/backend/blob/master/utility/user_reminder.py) script that will be triggered twice a year to show us what
 users we have in our database.
-* In the ietfYangDraftPull directory there are four jobs.
-    1. [draftPull](https://github.com/YangCatalog/backend/blob/master/ietfYangDraftPull/draftPull.py) adds new modules
-    to the YangModels/yang repository if there are any new modules. 
-    2. [draftPullLocal](https://github.com/YangCatalog/backend/blob/master/ietfYangDraftPull/draftPullLocal.py)
+* In the [ietfYangDraftPull](https://github.com/YangCatalog/backend/blob/master/ietfYangDraftPull) directory there are two jobs.
+    1. [draftPullLocal](https://github.com/YangCatalog/backend/blob/master/ietfYangDraftPull/draftPullLocal.py)
     goes through all ietf drafts and rfcs and populates yangcatalog.
```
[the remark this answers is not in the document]
richardzilincikPantheon: Also looks like `draftPullLocal` also populates iana modules, so renaming it would probably be helpful.

utility/script_config_dict.py
```diff
@@ -88,8 +89,12 @@ class ScriptConfigInfo(BaseScriptConfigInfo, total=False):
             },
         ],
     },
-    'ianaPull': {
-        'help': 'Pull the latest IANA-maintained files and add them to the Github if there are any new.',
+    'iana_push': {
+        'help': (
+            'Script automatically pushes new IANA-maintained yang modules to the GitHub repository:'
+            'https://github.com/yang-catalog/yang. Old ones are removed and their naming is corrected to '
+            '<name>@<revision>.yang.',
```
richardzilincikPantheon, line 96: This comma isn't supposed to be there, it turns it into a tuple.

automatic_push/README.md
```diff
@@ -25,3 +25,15 @@ This package contains python scripts to process IETF and openconfig YANG files:
     iana-if-type@2022-08-17.yang
     iana-if-type@2022-08-24.yang
     ```
+  
+- draft_push
+
+  Script automatically pushes new IETF draft yang modules to the GitHub repository: https://github.com/yang-catalog/yang.
+  Old ones are removed and their naming is corrected to ```<name>@<revision>.yang```. New IETF RFC modules are checked
```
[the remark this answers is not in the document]
SlavomirMazurPantheon: I would just add that if `draft_push.py` triggers `rfc_push.py` we should either rename them to `ietf_rfc_push.py` and `ietf_draft_push.py` OR create one script `ietf_push.py`, which will handle both, creating commits for RFC and Drafts and then pushing them to the `yang-catalog/yang`.

README.md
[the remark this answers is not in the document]
Fagtoy: I renamed that script to ```ietf_pull_local``` and changed any mentions to the new name, but do I understand correctly that this should not be changed: https://github.com/YangCatalog/backend/blob/develop/parseAndPopulate/groupings.py#L147?

automatic_push/README.md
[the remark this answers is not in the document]
Fagtoy: I moved functionality of ```ietf_rfc_push.py``` into ```ietf_push.py``` and removed the commit of only RFC files because they would be committed and pushed at the end of the script anyways within the cronjob, so there's no reason to distinguish them

utility/repoutil.py
```diff
@@ -103,30 +115,22 @@ def get_repo_owner(self) -> str:
...
-            :param config_username  (Optional[str]) Username to set in the git config.
-            :param config_user_email  (Optional[str]) Email to set in the git config.
+            :param clone_options  (Optional[CloneOptions]) Data for the repository cloning
         """
-        if local_dir:
+        if local_dir := clone_options.get('local_dir'):
             self.local_dir = local_dir
         else:
             self.local_dir = tempfile.mkdtemp()
         self.repo = Repo.clone_from(self.url, self.local_dir, multi_options=['--recurse-submodules'])
```
SlavomirMazurPantheon, line 130: I would leave this as something which is set by argument for example. We usually need to clone repo only because of IETF modules and rarely need to recursively pull all the submodules.
With submodules it takes too much space and time.

automatic_push/ietf_push.py
```diff
@@ -0,0 +1,206 @@
...
+        self.rfc_directory = os.path.join(ietf_directory, 'YANG-rfc')
+        self.is_production = self.config.get('General-Section', 'is-prod') == 'True'
+        log_directory = self.config.get('Directory-Section', 'logs')
+        self.logger = log.get_logger('ietf_push', f'{log_directory}/jobs/ietf-push.log')
+
+        self.repo: t.Optional[repoutil.ModifiableRepoUtil] = None
+
+    @job_log(file_basename=BASENAME)
+    def __call__(self):
+        self.logger.info('Starting Cron job IETF pull request')
```
SlavomirMazurPantheon, line 86: `Starting IETF push cronjob` or something like that

automatic_push/iana_push.py
```diff
@@ -0,0 +1,156 @@
...
+        self.iana_exceptions_file_path = self.config.get('Directory-Section', 'iana-exceptions')
+        self.is_production = self.config.get('General-Section', 'is-prod') == 'True'
+        log_directory = self.config.get('Directory-Section', 'logs')
+        self.logger = log.get_logger('iana_push', f'{log_directory}/jobs/iana-push.log')
+
+        self.repo: t.Optional[repoutil.ModifiableRepoUtil] = None
+
+    @job_log(file_basename=BASENAME)
+    def __call__(self):
+        self.logger.info('Starting job to pull IANA-maintained modules')
```
SlavomirMazurPantheon, line 73: `to push` instead of `to pull`?

automatic_push/utils.py
```diff
@@ -0,0 +1,166 @@
...
+from configparser import ConfigParser
+
+import requests
+from git import GitCommandError
+
+from utility import repoutil
+from utility.staticVariables import github_url
+
+
+def update_forked_repository(yang_models: str, forked_repo_url: str, logger: logging.Logger):
```
SlavomirMazurPantheon, line 37: As this method works with Github repos, cannot we move this functionality to repoutil.py?

automatic_push/utils.py
```diff
@@ -0,0 +1,166 @@
...
+                module_is_broken = '<html>' in lines[1] and '</html>' in lines[-1]
+            if module_is_broken:
+                logger.info(f'Deleted the file because of broken content: {key} - {yang_download_link}')
+                os.remove(file_path)
+            continue
+        with open(file_path, 'w') as yang_file:
+            yang_file.write(file_content_response.text)
+
+
+def push_untracked_files(
```
SlavomirMazurPantheon, line 118: Same for pushing, it would be better if all the functionality related to the manipulation with GitHub will be in same file.

README.md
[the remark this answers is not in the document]
richardzilincikPantheon: You can change it, but it needs to match up with this https://github.com/Fagtoy/backend/blob/2e01d819e351512b54ffd78ffdd633f348d05e7c/ietfYangDraftPull/pull_local.py#L123.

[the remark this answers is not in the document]
Fagtoy: okay, maybe it should be better to leave it as it is to not break anything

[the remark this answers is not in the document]
richardzilincikPantheon: It won't affect old schemas. It's just how we detected that the populate script was being run by draftPullLocal. That part of the path is cut off on the next line, so it doesn't show up in the resulting schema url.

automatic_push/utils.py
Fagtoy: I believe that this one is ```automatic_push``` directory specific, I moved pushing of all untracked files because we can use it in some other scripts in the future but not sure about these functions that are now in ```utils.py```

automatic_push/ietf_push.py
```diff
@@ -182,25 +173,62 @@ def _update_rfc_files_locally(self, new_files: list[str], diff_files: list[str])
         if self.send_message:
             self.logger.info('new or modified RFC files found. Sending an E-mail')
             mf = message_factory.MessageFactory()
-            if self.repo.repo.index.diff(None, paths=self.rfc_dir):
+            for _ in self.repo.repo.index.diff(None, paths=self.rfc_dir):
```
richardzilincikPantheon, line 176: why is this better than an if?

Fagtoy: ```repo.index.diff()``` returns a generator, so it's enough to check that at least one file is in diff

automatic_push/ietf_push.py
```diff
@@ -0,0 +1,234 @@
...
+        ietf_directory = self.config.get('Directory-Section', 'ietf-directory')
+        self.rfc_directory = os.path.join(ietf_directory, 'YANG-rfc')
+        self.is_production = self.config.get('General-Section', 'is-prod') == 'True'
+        log_directory = self.config.get('Directory-Section', 'logs')
+        self.logger = log.get_logger('ietf_push', f'{log_directory}/jobs/ietf-push.log')
+
+        self.repo: t.Optional[repoutil.ModifiableRepoUtil] = None
+
+    @job_log(file_basename=BASENAME)
+    def __call__(self):
```
SlavomirMazurPantheon, line 89: Is there any particular reason for using this approach with using `__call__` method?
I am not saying it is incorrect, it is not used anywhere in the solution.

Fagtoy: this class is only responsible for one job, so I thought it is reasonable to use this method here (and for IANA push as well)

richardzilincikPantheon (approved):

SlavomirMazurPantheon (approved):

PR #737, opened 2023-03-01
Fix organizations

### LINKED ISSUE #733: Fix organizations set for mef modules
Because ~~iana and~~ mef were missing from our predefined list of organizations, some modules that should have been assigned these organizations now have incorrect organizations.
We could use a script to reparse the organizations for all modules to fix this.
Since there are relatively few modules belonging to these organizations, perhaps this could also be done manually.

Fagtoy:
I removed IANA from organizations but left it in sdos because it was in sdos even before this PR: https://github.com/YangCatalog/backend/pull/729, I also searched for mef modules on PROD and it seems like there are no such modules with 'independent' organization, organizations set either to ietf or mef, should we change it to mef everywhere?

**NOTE**: In grep search, I assumed that modules names started with the organization name should have this organization (so ```iana-...``` modules have iana organization), but as it's not the case I removed this functionality

resolves #733


richardzilincikPantheon:
> organizations set either to ietf or mef, should we change it to mef everywhere?

Probably, it should be consistent in some way at least. Could you show an example where the organization is set to ietf?

richardzilincikPantheon:
> I removed IANA from organizations but left it in sdos because it was in sdos even before this PR

The list of SDOs is meant to be a subset of the organizations, just to check if a specific organization is an SDO. I think it's only used for ES. Since there are no modules with their organization set to iana, we won't ever need to check if iana is an SDO. It staying there won't make any functional difference.

Fagtoy:
> > organizations set either to ietf or mef, should we change it to mef everywhere?
> 
> Probably, it should be consistent in some way at least. Could you show an example where the organization is set to ietf?

Sorry, I wasn't right, I only found this module with ietf organization: https://yangcatalog.org/yang-search/module_details/ietf-l1-mef-service-types@2019-06-20, but it doesn't start with mef, so it's okay, it seems like all the mef modules have mef organization

Fagtoy:
> > organizations set either to ietf or mef, should we change it to mef everywhere?
> 
> Probably, it should be consistent in some way at least. Could you show an example where the organization is set to ietf?

Thanks, I removed iana from sdos

richardzilincikPantheon (approved):

SlavomirMazurPantheon (approved):

PR #738, opened 2023-03-03
chore: preffer SDO modules from official source

richardzilincikPantheon:


parseAndPopulate/groupings.py
```diff
@@ -116,10 +119,11 @@ def _parse_and_load_api(self) -> tuple[int, int]:
         self.logger.debug('Parsing sdo files sent via API')
         with open(os.path.join(self.dir_paths['json'], 'request-data.json'), 'r') as f:
             sdos_json = json.load(f)
-        sdos_list: t.List[dict] = sdos_json['modules']['module']
+        sdos_list: list[Module.AdditionalModuleInfo] = sdos_json['modules']['module']
         sdos_count = len(sdos_list)
         for i, sdo in enumerate(sdos_list, start=1):
             # remove diacritics
+            assert 'source-file' in sdo
```
Fagtoy, line 126: is it okay to raise an exception here? as I can see, we don't use try/except block in the ```parse_directory.py``` script for groupings so it would result in an actual error, maybe we should just use ```continue``` here if 'source-file' not in sdo? and assert isn't the best option if we really need to raise an exception here, I'd suggest using just KeyError here

parseAndPopulate/groupings.py
```diff
@@ -190,7 +194,14 @@ def _parse_and_load_not_api(self) -> tuple[int, int]:
...
                     self.log_module_creation_exception(e)
                     continue
+                if yang.organization != self.official_source and should_parse.seen_name_revision:
+                    # this is not the official source of this organization's modules
+                    # and we already have some version of this module
+                    continue
+                # this is the official source of this organization's modules
+                # or we don't have a version of this module yet
                 self.dumper.add_module(yang)
+                shutil.copy(path, self.file_mapping[path])
```
Fagtoy, line 204: shutil.copy(path, all_modules_path)

parseAndPopulate/groupings.py
```diff
@@ -272,7 +284,10 @@ def parse_and_load(self) -> tuple[int, int]:
             except (ParseException, FileNotFoundError) as e:
                 self.log_module_creation_exception(e)
                 continue
+            if yang.organization != self.official_source and should_parse.seen_name_revision:
+                continue
             self.dumper.add_module(yang)
+            shutil.copy(path, self.file_mapping[path])
```
Fagtoy, line 290: shutil.copy(path, all_modules_path)

parseAndPopulate/file_hasher.py
```diff
@@ -31,6 +31,12 @@
 BLOCK_SIZE = 65536  # The size of each read from the file
 
 
+@dataclass
+class SdoHashCheck:
+    seen_hash: bool
+    seen_name_revision: bool
```
Fagtoy, line 37: maybe:
```
hash_cahnged: bool
was_parsed_previously: bool
```
it would be more self-explanatory

parseAndPopulate/groupings.py
```diff
@@ -190,7 +194,14 @@ def _parse_and_load_not_api(self) -> tuple[int, int]:
                 except (ParseException, FileNotFoundError) as e:
                     self.log_module_creation_exception(e)
                     continue
+                if yang.organization != self.official_source and should_parse.seen_name_revision:
```
Fagtoy, line 197: shouldn't it be:
```
if self.official_source and yang.organization != self.official_source and should_parse.seen_name_revision:
```
as official_source can be None?

parseAndPopulate/groupings.py
```diff
@@ -272,7 +284,10 @@ def parse_and_load(self) -> tuple[int, int]:
             except (ParseException, FileNotFoundError) as e:
                 self.log_module_creation_exception(e)
                 continue
+            if yang.organization != self.official_source and should_parse.seen_name_revision:
```
Fagtoy, line 287: same as above

utility/script_config_dict.py
```diff
@@ -180,6 +180,12 @@ class ScriptConfigInfo(BaseScriptConfigInfo, total=False):
                 'type': str,
                 'default': '/var/yang/all_modules',
             },
+            {
+                'flag': '--official-source',
+                'help': 'If this is an official souce of an SDOs modules, set this option to the name of the SDO',
+                'type': str,
+                'default': 'None',
```
Fagtoy, line 187: 'None' as a string can be confusing, maybe just use  ```'default': '',```?

Fagtoy (changes requested):

parseAndPopulate/groupings.py
richardzilincikPantheon: At this point the dict is guaranteed to have a `'source-file'` key, this is checked by the api endpoint that that causes this script to be called (see [here](https://github.com/YangCatalog/backend/blob/3716c97031572f188513e54970493fa364eb2d5b/api/views/userSpecificModuleMaintenance/moduleMaintenance.py#L284)). The assert can not fail here, I inserted it to help the type checker. I should have included that information in the assert though, I'll fix that.

parseAndPopulate/groupings.py
richardzilincikPantheon: `self.official_source` can, but `yang.organization` has type `str` so it can't be `None`.

utility/script_config_dict.py
richardzilincikPantheon: Yep, that was unintentional.

parseAndPopulate/groupings.py
richardzilincikPantheon: Oh but that should be fine. The logic here is supposed to be:
- if we haven't seen the name@revision before, we accept it from anywhere (`should_parse.seen_name_revision` will be false)
- otherwise, we accept it only if the this is the official source for the organization of that module (so only if `yang.organization == self.official_source`)

If neither of these is the case, both of the conditions in the `if` evaluate to true and we skip the module.

Fagtoy: understood, thanks

Fagtoy (approved):

SlavomirMazurPantheon (approved):

PR #743, opened 2023-03-08
chore: add worktrees instead of recloning where possible

richardzilincikPantheon:


automatic_push/iana_push.py
```diff
@@ -66,17 +69,15 @@ def __init__(self, config: ConfigParser = create_config()):
         log_directory = self.config.get('Directory-Section', 'logs')
         self.logger = log.get_logger('iana_push', f'{log_directory}/jobs/iana-push.log')
 
-        self.repo: t.Optional[repoutil.ModifiableRepoUtil] = None
-
     @job_log(file_basename=BASENAME)
     def __call__(self) -> list[JobLogMessage]:
         self.logger.info('Starting job to push IANA-maintained modules')
-        self.repo = get_forked_repository(self.config, self.logger)
+        self.repo = get_forked_worktree(self.config, self.logger)
```
Fagtoy, line 75: shouldn't the repo be removed after executing the script as in the ```ietf_push.py``` repoutil.remove_worktree(self.repo.working_dir)?

automatic_push/utils.py
```diff
@@ -16,79 +16,249 @@
...
-from utility import repoutil
+from utility import repoutil, yangParser
 from utility.staticVariables import github_url
+from utility.util import revision_to_date
 
 
-def get_forked_repository(config: ConfigParser, logger: logging.Logger) -> repoutil.ModifiableRepoUtil:
+def get_forked_worktree(config: ConfigParser, logger: logging.Logger) -> Repo:
     """
     Returns the ModifiableRepoUtil instance of the https://github.com/yang-catalog/yang repository
```
Fagtoy, line 39: it doesn't return the ModifiableRepoUtil anymore

automatic_push/utils.py
```diff
@@ -16,79 +16,249 @@
...
+          :param logger (logging.Logger) Logger instance to log information/exceptions.
+          :param verified_commits_file_path (str) Path to file where our verified commits should be stored
+          in order to verify commits in GitHub webhooks.
+          :param is_production (bool) If set to False then files would only be committed and not pushed to the repo.
+    :return (PushResult) Returns information about the push.
+    """
+    try:
+        untracked_files = repo.untracked_files
+        logger.info('Committing all files locally')
+        repo.git.commit(a=True, m=commit_message)
```
Fagtoy, line 240: [The -a option only stages changes to files that are already tracked in the repository. Any new files that you've added to the repository will still need to be explicitly staged using git add before they can be committed](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--a), so there should be a ```git add .``` command first

automatic_push/utils.py
```diff
@@ -16,79 +16,249 @@
...
+        logger.info(f'Commit hash {commit_hash}')
+        with open(verified_commits_file_path, 'w') as f:
+            f.write(f'{commit_hash}\n')
+        if is_production:
+            logger.info('Pushing untracked and modified files to remote repository')
+            repo.git.push('fork')
+        else:
+            logger.info('DEV environment - not pushing changes into remote repository')
+            untracked_files_list = '\n'.join(untracked_files)
+            logger.debug(f'List of all untracked and modified files:\n{untracked_files_list}')
```
Fagtoy, line 252: I just realized that there will be no untracked changes at this point, so if I'm not mistaken, all the modified/deleted/new files can be listed using this iterator ```repo.index.diff(None)```

ietfYangDraftPull/pull_local.py
```diff
@@ -27,28 +27,40 @@
...
+        return self
+
+    def __exit__(self, *exc_info):
+        for worktree in self._worktrees.values():
+            remove_worktree(worktree)
+
+    def __getitem__(self, repo_dir: str):
+        if repo_dir not in self._worktrees:
+            self._worktrees[repo_dir] = add_worktree(repo_dir)
+        return self._worktrees[repo_dir]
```
Fagtoy, line 61: it seems like a class more related to the repoutil, don't you think so?

ietfYangDraftPull/pull_local.py
```diff
@@ -107,63 +110,32 @@ def main(script_conf: ScriptConfig = DEFAULT_SCRIPT_CONFIG.copy()) -> list[JobLo
...
-
-        directory_success, message = populate_directory(experimental_path, notify_indexing, logger)
-        success = success and directory_success
-        messages.append({'label': 'Experimental modules', 'message': message})
-
-        # IANA modules
-        iana_path = os.path.join(repo.local_dir, 'standard/iana')
-
-        if os.path.exists(iana_path):
-            directory_success, message = populate_directory(iana_path, notify_indexing, logger)
```
Fagtoy, line 160: what about all this specific logic, isn't it needed anymore?

utility/repoutil.py
```diff
@@ -363,6 +252,7 @@ class PullRequestApprovingDetail(t.TypedDict, total=False):
     commit_id: str
     body: str
     comments: list[dict]
+    event: str
```
Fagtoy, line 255: the event is set automatically to 'APPROVE' anyway

utility/repoutil.py
```diff
@@ -315,6 +202,8 @@ class PullRequestCreationDetail(t.TypedDict, total=False):
     maintainer_can_modify: bool
     draft: bool
     issue: int
+    head: str
+    base: str
```
Fagtoy, line 206: these fields are set using the positional arguments ```head_branch``` and ```base_branch``` in the ```create_pull_request``` function itself

ietfYangDraftPull/pull_local.py
```diff
@@ -107,63 +110,32 @@ def main(script_conf: ScriptConfig = DEFAULT_SCRIPT_CONFIG.copy()) -> list[JobLo
...
-        logger.exception('Exception found while running pull_local script')
-        raise e
+
+    with WorktreeManager() as worktrees:
+        try:
+            for repo, path_in_repo, label in (
+                (yang_models_dir, 'standard/ietf/RFC', 'Standard RFC modules'),
+                (yang_models_dir, 'experimental/ietf-extracted-YANG-modules', 'Experimental modules'),
+                (yang_models_dir, 'standard/iana', 'IANA modules'),
+                (openconfig_dir, '', 'OpenConfig modules'),
```
Fagtoy, line 130: what about specific logic for populating openconfig modules, is it just removed?

utility/repoutil.py
```diff
@@ -139,6 +97,7 @@ def _clone(self, clone_options: CloneOptions):
                 config.set_value('user', 'name', config_username)
 
 
+# This probably isn't worth it's own class anymore
 class ModifiableRepoUtil(RepoUtil):
```
Fagtoy, line 101: I believe that this class and the RepoUtil class should be removed if they aren't used anymore

Fagtoy (changes requested):
And also, all mentions of the ```openconfigPullLocal.py``` script should be removed as well (documentation, receiver, crontab, launch.json, etc.)

automatic_push/iana_push.py
richardzilincikPantheon: it is, on line 96

ietfYangDraftPull/pull_local.py
richardzilincikPantheon: As far as I can tell, the logic was separate only because we wanted to clone the openconfig repo every time. We no longer need to do that if we keep the openconfig repo in `/var/yang/nonietf` up to date, so there is no need for extra logic/

utility/repoutil.py
richardzilincikPantheon: Yeah your right, I forgot this comment there.

ietfYangDraftPull/pull_local.py
Fagtoy: not sure, for example, this part:
```
 response = requests.get(ietf_rfc_url)
        tgz_path = os.path.join(repo.local_dir, 'rfc.tgz')
        extract_to = os.path.join(repo.local_dir, 'standard/ietf/RFC')
        with open(tgz_path, 'wb') as zfile:
            zfile.write(response.content)
        tar_opened = draftPullUtility.extract_rfc_tgz(tgz_path, extract_to, logger)
```
makes sense because not all modules can be already available in the ```standard/ietf/RFC``` dir (the PR from us isn't automatically merged yet, for example), so we make sure that we populate only the latest modules using the ```ietf_rfc_url```

Fagtoy: it depends on when the script runs, for now, the ```ietf_push.py``` script runs at 5 14 and ```pull_local.py``` runs at 5 22, it's not enough time to finish the workflows in the ```yang-catalog``` and in the ```YangModels``` repos

utility/repoutil.py
Fagtoy: I understand that there can be a warning, but it's not that critical in my opinion, adding these args to the ```PullRequestCreationDetail``` isn't good, because we can't be sure that these args are actually provided, and raising an error here will not really help us, as the solution I would prefer doing something like: 
```
request_body = request_body or PullRequestCreationDetail()
request_body = request_body.copy()
request_body['head'] = head_branch
request_body['base'] = base_branch
```
as far as I understand, it shouldn't highlight any warnings

ietfYangDraftPull/pull_local.py
richardzilincikPantheon: I see. Looking at past runs of the workflows, it looks like they both generally finish within 1 hour and 30 min. If we move `pull_local.py` to run 2 hours after `ietf_push.py` we should be fine, right?

Fagtoy: right, +-4 hours (~1.5h for fork + ~1.5h for YangModels + some delta), but I wouldn't consider this method a safe one for the future, to be honest, maybe it needs at least mentioned somewhere, if workflows take more time we will populate new/changed modules the next day, not sure if it's critical. Maybe we can use the strategy pattern here in some way and for specific worktrees use specific population strategies, which will in this example just update the modules in the right way, so our code will be more scalable for the future as well

utility/repoutil.py
richardzilincikPantheon: I agree that it's not critical in this specific case, but the type checker does catch important errors and it's easier to spot them when there aren't other unimportant warnings in the file.

I found an approach that works gets rid of the error.
```
request_body_dict = dict(request_body or PullRequestCreationDetail())
request_body_dict['head'] = head_branch
request_body_dict['base'] = base_branch
```

ietfYangDraftPull/pull_local.py
richardzilincikPantheon: The fork actually only triggers the check for SDO modules, which runs in less than 10 mins https://github.com/yang-catalog/yang/actions. I just thought of something though, we could just run the populate script over the `/var/yang/ietf/YANG-rfc` directory if we don't want to wait for the repo.

Fagtoy: Yeah, basically, if that dir includes all the needed modules (not sure about this) then there’s no need to use the repo at all

richardzilincikPantheon: I believe the directories are both supposed to have the same contents, with the addition of the symlinks without revisions in standard/ietf/RFC. However it seem there are some differences
```
1,3d0
< example-dhcpv6-class-select@2022-06-20.yang
< example-dhcpv6-opt-sip-serv@2022-06-20.yang
< example-dhcpv6-server-conf@2022-06-20.yang
10a8
> iana-if-type@2017-01-19.yang
12a11
> iana-routing-types@2018-10-29.yang
31a31
> ietf-connectionless-oam-methods@2019-04-16.yang
132a133
> ietf-restconf-monitoring@2017-01-26.yang
134c135
< ietf-rip@2020-02-20.yang
---
> ietf-restconf@2017-01-26.yang
184a186
> ietf-yang-structure-ext@2020-06-17.yang
```
I think this is unintentional so I'll try and figure out what's causing this.

richardzilincikPantheon: So the modules which are present in `/var/yang/ietf/YANG-rfc` but not in `standard/ietf/RFC` are blacklisted in the `/var/yang/ietf-exceptions/exceptions.dat` file to prevent them from being pushed to the repo. This would still not happen in the pull script, so it's not an issue.

These modules
```
iana-if-type@2017-01-19.yang
iana-routing-types@2018-10-29.yang
ietf-connectionless-oam-methods@2019-04-16.yang
```
seem to have been added to `standard/ietf/RFC` manually by accident.

And the last three
```
ietf-restconf-monitoring@2017-01-26.yang
ietf-restconf@2017-01-26.yang
ietf-yang-structure-ext@2020-06-17.yang
```
aren't getting extracted anymore because of a xym error.
```
Extracting 'ietf-restconf'
   ERROR: '/var/yang/ietf/rfc/rfc8040.txt', Line 4761 - 'module' statement within another module
```

automatic_push/ietf_push.py
```diff
@@ -173,7 +178,7 @@ def _update_rfc_files_locally(self, new_files: list[str], diff_files: list[str])
         if self.send_message:
             self.logger.info('new or modified RFC files found. Sending an E-mail')
             mf = message_factory.MessageFactory()
-            if self.repo.repo.untracked_files or self.repo.repo.index.diff(None):
+            for _ in self.repo.index.diff(None, paths=self.rfc_dir):
```
Fagtoy, line 181: maybe you didn't pull the latest changes from the develop but it should be: ```if self.repo.untracked_files or self.repo.index.diff(None):```

Fagtoy (approved):

PR #746, opened 2023-03-10
fix: fixed remove_unused.py script

### LINKED ISSUE #745: Fix ```remove_unused.py``` script
During runs of the ```remove_unused.py``` script, there are situations when the ```PermissionDenied``` error is raised, it should be wrapped in the try/except block

Fagtoy:
I couldn't find the root cause of this issue locally, so I just wrapped it in the try/except block, and I also refactored the script a little bit

resolves #745

### DIFF
--- a/utility/remove_unused.py
+++ b/utility/remove_unused.py
@@ -28,6 +28,7 @@ import os
 import shutil
 import time
 import uuid
+from datetime import datetime
 from datetime import datetime as dt
 
 import utility.log as log
@@ -40,22 +41,6 @@ DAY = 86400
 BLOCK_SIZE = 65536
 
 current_file_basename = os.path.basename(__file__)
-
-
-def represents_int(s: str) -> bool:
-    try:
-        int(s)
-        return True
-    except ValueError:
-        return False
-
-
-def represents_uuid(s: str) -> bool:
-    try:
-        uuid.UUID(s)
-        return True
-    except ValueError:
-        return False
 
 
 @job_log(file_basename=current_file_basename)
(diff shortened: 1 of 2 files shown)


utility/remove_unused.py
```diff
@@ -136,97 +126,109 @@ def main():
...
+
+def remove_old_backups(backup_directory: str):
+    """
+    Deduplicate backups and keep the latest copy only. Keep one backup per month for the last 6 months.
+    For the last 2 month, keep all unique backups. All other files are removed.
+
+    Arguments:
+        :param backup_directory  (str) Path to directory to search for backups.
+    """
+    list_of_backups = get_list_of_backups(backup_directory)
```
SlavomirMazurPantheon, line 161: In `get_list_of_backups` please handle the case if directory does not exists.
It will fail on `os.listdir(dir_name)` now

utility/remove_unused.py
```diff
@@ -136,97 +126,109 @@ def main():
...
+    backup_name_latest = os.path.join(backup_directory, list_of_backups[-1])
```
SlavomirMazurPantheon, line 162: I've got following exception:
```

  File "/backend/utility/remove_unused.py", line 130, in main
    remove_old_backups(os.path.join(cache_directory, 'confd'))
  File "/backend/utility/remove_unused.py", line 162, in remove_old_backups
    backup_name_latest = os.path.join(backup_directory, list_of_backups[-1])
IndexError: list index out of range
```
Probably the case if directory does not exist at all or does not contain any files.

SlavomirMazurPantheon (changes requested):

PR #767, opened 2023-05-03
chore: remove sender/receiver

richardzilincikPantheon:


api/views/user_specific_module_maintenance.py
```diff
@@ -592,3 +573,23 @@ def get_repo(repo_url: str, owner: str, repo_name: str) -> RepoUtil:
                 description='bad request - could not clone the Github repository. Please check owner,'
                 ' repository and path of the request - {}'.format(e.stderr),
             )
+
+
+def get_response(correlation_id: str) -> str:
```
Fagtoy, line 578: I believe this function would better be placed in the `Job_runner.py` because this functionality isn't limited to only `user_specific_module_maintenance.py`

Fagtoy (changes requested):
Don't you think that writing job statuses to separate files would be better, so we can see if there were some jobs running while the container was rebuilt for example? And maybe it's also worth it to set the max workers of the Pool to 1, so we don't get inconsistent results from different scripts running simultaneously, or if the same script is being executed twice at the same time?

api/views/user_specific_module_maintenance.py
richardzilincikPantheon: Are there any other places where it could be useful?

richardzilincikPantheon:
> Don't you think that writing job statuses to separate files would be better, so we can see if there were some jobs running while the container was rebuilt for example?

There's already logging in some of methods of the job runner. Would it be fine to just add logging to the places it's missing?

richardzilincikPantheon:
I found out that this won't work correctly with creating a pool in the server process. I'll implement running the processes differently.

api/views/user_specific_module_maintenance.py
Fagtoy: I mean in the future it's possible to see the status of a job, so Maybe in the future, we will need it somewhere else, and it's not obvious that this functionality is placed in the `user_specific_module_maintenance.py` module, I would expect it to be placed in the same module as the runner itself

Fagtoy:
> > Don't you think that writing job statuses to separate files would be better, so we can see if there were some jobs running while the container was rebuilt for example?
> 
> There's already logging in some of methods of the job runner. Would it be fine to just add logging to the places it's missing?

I think yes, I believe it would be better to add logging to the missing places

richardzilincikPantheon:
So after more investigation, it looks like a generally bad idea to use multiprocessing when using gunicorn. I didn't realize this before because the tests don't actually run against gunicorn.
There is however a better way of doing things than what we have now, which is to use [Celery](https://flask.palletsprojects.com/en/2.3.x/patterns/celery/) to manage the communication with the message broker. I'll close this PR and create a separate one using Celery.

PR #769, opened 2023-05-08
fix: fixed automatic pushes

Fagtoy:
Added pushing to the 'main' branch from the 'fork-main' branch at the end of the `ietf_push.py` and `iana_push.py` scripts


dmytro-kyrychenko:
resolves #778

PR #783, opened 2023-06-08
Redis databases code readability

dmytro-kyrychenko:
Also need to add this line to config:
redis-dir=${var}/redis

Also add such json file to this redis-dir folder, introduced above:

redis_databases.json 
```
{
  "SECRETS": 0,
  "MODULES": 1,
  "USERS": 2,
  "VENDORS": 4,
  "TEMP_MODULES": 5,
  "TEST_MODULES": 6,
  "USERS_NOTIFICATIONS": 7,
  "TEST_VENDORS": 9
}
```


redisConnections/redisConnection.py
```diff
@@ -42,12 +43,12 @@ def __init__(
         self._redis_host = config.get('DB-Section', 'redis-host')
         self._redis_port = int(config.get('DB-Section', 'redis-port'))
         if modules_db is None:
-            modules_db = config.get('DB-Section', 'redis-modules-db', fallback=1)
+            modules_db = config.get('DB-Section', 'redis-modules-db', fallback=RedisEnum.MODULES.value)
```
richardzilincikPantheon, line 46: Using two separate config files here is redundant. `redis-modules-db` never got added to the sample config or to the config in PROD afaik, so I think we should just get rid of these `config.get`s and just use the `RedisEnum` values directly.

richardzilincikPantheon (changes requested):

redisConnections/redisConnection.py
dmytro-kyrychenko: I think they can be used in some tests, cause we have them in test.conf for backend:
```
redis-modules-db=11
redis-vendors-db=14
redis-users-db=12
```

richardzilincikPantheon: Could these tests be changed to use the `TEST_MODULES`, etc. databases instead?

dmytro-kyrychenko: Yeah, I think I can change it right now

dmytro-kyrychenko: Actually, I checked and I am now not sure if we want to do this, because we have tests that are not using RedisConnection or RedisUsersConnection directly (these are the only classes that use redis-modules-db, redis-vendors-db and redis-users-db), but they are testing other code, that uses default variant of this classes and basically if we want to remove those config.get()'s, then we need to mock every such test (we need to add test parameters for every default RedisConnection and RedisUsersConnection)
So, do we still want to do this?

richardzilincikPantheon: I think we should do something similar to what we do with the .conf file. So have a file in `/var/yang/redis` or whatever with the real values, and have a file in `tests/resources` with test values.
So we won't need a `TEST_MODULES` entry, we'll just use the `MODULES` entry in the test config while testing and the `MODULES` entry in the real config when actually running. Does that make sense?

dmytro-kyrychenko: Yeah, totally, that will work

tests/resources/redis_databases.json
```diff
@@ -0,0 +1,11 @@
+{
+  "SECRETS": 0,
+  "MODULES": 11,
+  "USERS": 12,
+  "VENDORS": 14,
+  "TEMP_MODULES": 5,
+  "TEST_MODULES": 6,
```
richardzilincikPantheon, line 7: Can we get rid of the separate `TEST_MODULES` database now and use db 11 for all the tests (same for vendors)?

SlavomirMazurPantheon (approved):

dmytro-kyrychenko:
#797

PR #795, opened 2023-08-03
Refactor test dirs

dmytro-kyrychenko:


.pre-commit-config.yaml
```diff
@@ -1,19 +1,19 @@
...
       - id: black
   - repo: https://github.com/pycqa/isort
-    rev: 5.10.1
+    rev: 5.12.0
     hooks:
       - id: isort
         name: isort (python)
   - repo: https://github.com/pycqa/flake8
-    rev: '5.0.4'
+    rev: '6.1.0'
```
SlavomirMazurPantheon, line 16: @dmytro-kyrychenko Please move changes to this file into separate pull request.
Question: Is there any reason why this version is in quotes and others are not? If not please unify it.

tests/test_opensearch_manager.py
```diff
@@ -36,15 +35,14 @@ class TestOpenSearchManagerClass(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
-        config = create_config()
         opensearch_host_config = {
-            'host': config.get('DB-Section', 'opensearch-host', fallback='localhost'),
-            'port': config.get('DB-Section', 'opensearch-port', fallback='9200'),
+            'host': 'yc-opensearch',
```
SlavomirMazurPantheon, line 39: @dmytro-kyrychenko All OpenSearch tests are failing because of connection error. `yc-opensearch` is not correct value for tests which are running in Github Actions.
In `test.conf` file you can see that `opensearch-host` should have value `localhost`.
Basically you should revert this line as it most probably cause the harm.

.pre-commit-config.yaml
dmytro-kyrychenko: okay, I will create separate pr for this
I am not sure why it is in quotes since we don't actually manually change this files, pre-commit system does this automatically
Should I still change this in new PR?

tests/test_opensearch_manager.py
dmytro-kyrychenko: About this, I removed this line for the tests, because in this file for tests for opensearch we use real value of running opensearch (yc-opensearch locally), and in all other files, which also use value from config it is set to localhost, which I think is then replaced by mocking
So the config value for this file should be different than for every other test file
I am not sure how to solve this

dmytro-kyrychenko:
for now I will leave changes in pre-commit config file also here (I've created separate PR for this here - https://github.com/YangCatalog/backend/pull/796), since without them the whole pre-commit system doesn't work
I'll remove them from here, when the other PR is merged to develop

SlavomirMazurPantheon:
> for now I will leave changes in pre-commit config file also here (I've created separate PR for this here - #796), since without them the whole pre-commit system doesn't work I'll remove them from here, when the other PR is merged to develop

PR merged

tests/test_opensearch_manager.py
SlavomirMazurPantheon: @dmytro-kyrychenko I would still revert this back so we can successfully run the tests. Rather change it locally when you will run the tests.

PR #816, opened 2024-02-17
Restore the .json extension on ytree files.

xorrkaz:
This was removed with os.path.join() was introduced.  Other parts of the code expect the files to end in .json.  Without this the JS-based YANG trees are broken.


microamp:
Already patched in prod. The failing checks have been addressed in #819. However, #819 has not been released yet, so it will remain open until all PRs that have been patched in prod are merged first.

PR #817, opened 2024-02-17
Remove legacy IP addresses.

xorrkaz:
Keeping IPs in the code is too fragile.  Remove them and only validate the hostnames.


microamp:
Already patched in prod. The failing checks have been addressed in https://github.com/YangCatalog/backend/pull/819. However, https://github.com/YangCatalog/backend/pull/819 has not been released yet, so it will remain open until all PRs that have been patched in prod are merged first.

PR #820, opened 2024-02-23
Ignore a missing module revision.

xorrkaz:
Currently, if a module's revision shows up in the index (but not redis) a search for the module's details will fail with a 404, even if other revisions of the same module _do_ show up.  Be a little more permissive here.  If the module revision that is desired (or the latest) isn't found, then return 404.  Else, log the condition and continue processing revisions.


microamp:
Running in prod

PR #822, opened 2024-02-27
ci: add semantic check config

microamp:
Adding a checker for [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), which is also used in `ietf-tools` repos 

e.g. https://github.com/ietf-tools/datatracker/blob/main/.github/semantic.yml

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,0 +1,3 @@
+enabled: true
+titleOnly: true
+targetUrl: "https://www.conventionalcommits.org/en/v1.0.0/#summary"


NGPixel (approved):

microamp:
> This won't do anything until we add the Semantic check app on the github organization though.

Thanks @NGPixel. I was wondering about that - will look into it.

microamp:
Update: Amended the existing commit with no changes to confirm that Semantic PR correctly triggers.

PR #826, opened 2024-03-05
chore: Ignore Q000 in flake8 config

microamp:
~Use single quotes for Python strings to fix the flake8 error.~

```text
./parseAndPopulate/modulesComplicatedAlgorithms.py:432:34: Q000 Double quotes found but single quotes preferred
Error: Process completed with exit code 1.
```

**Update**: `flake8` doesn't seem to support `black` as profile, like `isort` does. Going forward, we'll choose `black`'s formatting over `flake8`'s when there is a conflict between the two.


microamp:
Update: All checks now pass.

